Issue #364, opened 2021-10-15
CSS/HTML fix for top bar in documentation


edublancas:
For certain screen width (shorter than a monitor but larger than a phone), the top bar in the docs gets weird:

![Screen Shot 2021-10-15 at 11 43 33](https://user-images.githubusercontent.com/989250/137515655-cce817cc-20ce-4ddd-a2e0-ea5eae3af2a0.png)

Source code: https://github.com/ploomber/ploomber/blob/master/doc/_templates/macros.html

We're using [bootstrap](https://getbootstrap.com/)

edublancas:
let me work on pushing instructions for building the docs locally, will update you once it's up

edublancas:
done, feel free to reach out if you have questions 

https://github.com/ploomber/ploomber/blob/master/doc/CONTRIBUTING.md

rehman000:
Okay @idomic I'll look into this as well. Thanks for the heads up! 👍

PR #374, opened 2021-10-16
Fixed Responsive bootstrap navbar issue #364

rehman000:
- Also added instructions on how to get into the virtual environment on windows.

### DIFF
--- a/doc/_templates/macros.html
+++ b/doc/_templates/macros.html
@@ -1,6 +1,6 @@
 {% macro nav_bar(pathto, active=none) %}
 
-<nav class="navbar navbar-expand-md navbar-dark bg-dark fixed-top">
+<nav class="navbar navbar-expand-md navbar-dark fixed-top">
     <button class="navbar-toggler" type="button" data-toggle="collapse" data-target="#first-column"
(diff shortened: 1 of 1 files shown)


rehman000:
Question @edublancas why is the directory that contains `index.html`  in `.gitignore`?

image

No wonder my changes are not being committed. Is there any other file that I needed to add these changes to? Like `macros.html` or was it some other file. Given `index.html` inside the `/doc/_build/` directory will be ignored by git.

edublancas:
yes, _build is ignored because that's the final output of the doc building process. please apply your changes to the doc source code.

in your case, the nav bar definition is here https://github.com/ploomber/ploomber/blob/master/doc/_templates/macros.html

rehman000:
Sorry I'm a bit unfamiliar with the project directory structure. I apologize.

edublancas:
no worries!

rehman000:
On localhost, This is what I see for `macros.html` I can't even see a navbar. Did I mess something up? I used conda rather than pip (I doubt that matters). 

image

edublancas:
Awesome! Merging now, thanks a lot!

Issue #373, opened 2021-10-16
Add a function to find cells with the requested tags


edublancas:
This is a simple, isolated issue.

We have a function to iterate over cells to find one with the given tag:

https://github.com/ploomber/ploomber/blob/3b7997d88a34f75855c0b5ec4391c82dce4213b6/src/ploomber/sources/nb_utils.py#L1

However, this function has to clean up all cells with the given cells:

https://github.com/ploomber/ploomber/blob/3b7997d88a34f75855c0b5ec4391c82dce4213b6/src/ploomber/sources/notebooksource.py#L556

We're currently iterating the notebook one per tag but it'd be better to search for all tags on a single pass.

For example, if my input is:

```python
tags = ['a', 'b', 'c']
```

After running the function, the output should return a dictionary with the tags and its corresponding index:

```python
# 'a' is in the first cell, etc...
result = {'a': 0, 'b': 10, 'c': 20}
```

Note that cells may have more than one tag, but as long as it contains the one we are looking for, we don't care about the rest.

If the cell is not in the notebook, it should not appear in the output. Following our previous example:

```python
# input
tags = ['a', 'b', 'c']
```

Output:

```python
# if we only get 'a', it means 'b' and 'c' are not in the notebook
result = {'a': 0}
```

[I recently wrote something like this](https://github.com/edublancas/jupyblog/blob/cf2063007915b1c2d9519cd365e494075c584059/src/jupyblog/md.py#L44), except it looks for lines in a text file, as opposed to cells in a notebook, but the general logic remains the same. 

Tasks:

- [ ] Write a function that supports multiple tags
- [ ] Have `_cleanup_rendered_nb` use that new function

edublancas:
Sure! thanks for your interest!

please check out the contributing guide for setup instructions https://github.com/ploomber/ploomber/blob/master/CONTRIBUTING.md

and feel free to ping us on slack: http://community.ploomber.io/

edublancas:
thanks for your interest!

PR #385, opened 2021-10-18
add: find cells from multiple tags

Harsh-1309:
Hi @edublancas! I tried to complete the first task of implementing function which supports multiple tags.
What the current function does:
* It will take any number of tags as parameters.
* If a cell is found with any of these tags, it'll be returned along with it's index.

Let me know if any other change is required such as:
* If there are n-tags passed as parameters, then a cell which has all these n tags should be returned.


edublancas:
Hi! Thanks a lot for your work! Apologies, I think my initial description wasn't clear enough. Please refer to the updated description in the original issue. let me know if you still have time to make the changes and if there are any questions https://github.com/ploomber/ploomber/issues/373

PR #410, opened 2021-10-30
Improved error message when dotted path fails to load. Also, added respective tests for the same.

### LINKED ISSUE #405: Improve error message when dotted path fails to load
The `load_dotted_path` raises the  following error if unable to load the module:

```pytb
  File "/Users/Edu/Desktop/import-error/script.py", line 4, in <module>
    load_dotted_path('tests.quality.fn')
  File "/Users/Edu/dev/ploomber/src/ploomber/util/dotted_path.py", line 128, in load_dotted_path
    module = importlib.import_module(mod)
  File "/Users/Edu/miniconda3/envs/ploomber/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 984, in _find_and_load_unlocked
ModuleNotFoundError: An error happened when trying to import dotted path "tests.quality.fn": No module named 'tests.quality'
```

I ran into this issue when trying to import from a local `tests/quality.py` script because IPython registers the `tests` module:

```python
import tests

tests
<module 'tests' from '/Users/Edu/miniconda3/envs/ploomber/lib/python3.9/site-packages/IPython/extensions/tests/__init__.py'>
```

Hence, it shadows my module. It'd be better if we provided extra information to the user. For example, if loading `tests.quality.fn` fails, we could check if we can locate `tests`, and included the path to it as part of the error message, so the error becomes something like this:

```pytb
ModuleNotFoundError: An error happened when trying to import dotted path "tests.quality.fn": No module named 'tests.quality' (loaded 'tests' module from /patht/to/tests/__init__.py )
```

To hint to the user that the problem is that Python is loading the incorrect module, we can use something like this:

```python
# if tests.quality.fn fails, try importing just tests, if it works, show the origin as part of the error message
importlib.util.find_spec('tests').origin
```

Definition is here:

https://github.com/ploomber/ploomber/blob/fd9b4c7a2e787c0206f841928d1be90ac142c7a8/src/ploomber/util/dotted_path.py#L107

Tasks:

- [ ] fix load_dotted_path
- [ ] add tests to https://github.com/ploomber/ploomber/blob/master/tests/util/test_dotted_path.py

abhishak3:
Included path of the module in the error message imported through the `load_dotted_path` method. Also, added tests for the same.

Resolves #405


edublancas:
adding the tag so it counts for hacktoberfest :) will review later

abhishak3:
Thanks a lot !

edublancas:
thanks @abhishak3!

Issue #357, opened 2021-10-14
Improve error message when failing to initialize Metaproduct


edublancas:
Tasks may generate more than one product like this:

```python
from ploomber.products import File
from ploomber.tasks import PythonCallable
from ploomber import DAG


def _do_stuff():
    pass

# note we are calling FIle
PythonCallable(_do_stuff, {'a': File('something')}, dag=DAG())
```

But if the user forgets that:

```python
# forgot to call File!
PythonCallable(_do_stuff, {'a': 'something'}, dag=DAG())
```

We get this error:

```pytb
~/dev/ploomber/src/ploomber/tasks/tasks.py in __init__(self, source, product, dag, name, params, unserializer, serializer)
    100         self._source = type(self)._init_source(source, kwargs)
    101         self._unserializer = unserializer or dag.unserializer
--> 102         super().__init__(product, dag, name, params)
    103
    104     @staticmethod

~/dev/ploomber/src/ploomber/tasks/abc.py in __init__(self, product, dag, name, params)
    192             type(self).__name__))
    193
--> 194         self.product.task = self
    195         self._client = None
    196

~/dev/ploomber/src/ploomber/products/metaproduct.py in task(self, value)
    113     def task(self, value):
    114         for p in self.products:
--> 115             p.task = value
    116
    117     def exists(self):

AttributeError: 'str' object has no attribute 'task'
```

Better: check if `p` has a task attribute, if it doesn't, raise a better error like "doesn't look like a Product instance, got object of type {type}"

lfunderburk:
Can I have this assigned please?

idomic:
In addition, you can also join the slack channel where we have the Ploomber community and supporters: http://community.ploomber.io/

idomic:
Closing since the PR was merged.
Great job @lfunderburk !

PR #412, opened 2021-11-02
Catch error when p does not have attribute Task

lgfunderburk:
Contributes to 

https://github.com/ploomber/ploomber/issues/357 

Summary: modified metaproduct.py

```
def task(self, value):
    for p in self.products:
        if hasattr(p, 'task'):
            p.task = value
            continue
        else:
            raise TypeError(f"It doesn't look like a Product instance, got object of type {type(p)}")

```

currently working on adding test to https://github.com/ploomber/ploomber/blob/master/tests/products/test_metaproduct.py

### DIFF
--- a/src/ploomber/products/metaproduct.py
+++ b/src/ploomber/products/metaproduct.py
@@ -112,7 +112,11 @@ class MetaProduct(Mapping):
     @task.setter
     def task(self, value):
         for p in self.products:
-            p.task = value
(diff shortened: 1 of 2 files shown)


idomic:
@lgfunderburk is this ready for review?

lgfunderburk:
Yes - review sounds good

edublancas:
Thanks, @lgfunderburk for your contribution!

PR #436, opened 2021-12-11
Scaffold functions implement scaffolding

fferegrino:


tests/util/test_dotted_path.py
```diff
@@ -402,3 +403,24 @@ def fn(some_arg):
...
+    assert Path(tmp_directory, "sweet", "home", "__init__.py").exists()
+    assert Path(tmp_directory, "sweet", "home", "alabama.py").exists()
+
+
+def test_create_intermediate_modules_single(tmp_directory):
+    modules_and_function = ["ploomber"]
+
+    create_intermediate_modules(modules_and_function)
+
+    assert Path(tmp_directory, "ploomber.py").exists()
```
fferegrino, line 426: We can probably combine these tests if we sacrifice some readability.

edublancas:
Found an edge case, if a user does not pass an explicit path to a `pipeline.yaml`, scaffold calls `load_dag`:

https://github.com/ploomber/ploomber/pull/436/commits/928c5aaa1fd098eee9f8585b233559c536601bcb

Can you add a few more test cases? You can parameterize your existing tests, but do not pass the `['-e', 'pipeline.serve.yaml']` args so ploomber looks for the default `pipeline.yaml` file and it goes into this conditional:

https://github.com/ploomber/ploomber/blob/ecc5d0346db3b415c3845025aafcf2c8aff08e90/src/ploomber/cli/cli.py#L60

tests/cli/test_scaffold.py
```diff
@@ -212,6 +213,72 @@ def test_scaffold_in_custom_entry_point(tmp_directory):
...
+    assert module_file.exists() == create_module
+    Path('pipeline.serve.yaml').write_text("""
+tasks:
+    - source: my_module.my_function
+      product: out.ipynb
+""")
+
+    runner = CliRunner()
+    result = runner.invoke(scaffold,
+                           args=['-e', 'pipeline.serve.yaml'],
```
edublancas, line 233: please parameterize this test case, one should not pass any args so ploomber uses the default pipeline.yaml

tests/cli/test_scaffold.py
```diff
@@ -212,6 +213,72 @@ def test_scaffold_in_custom_entry_point(tmp_directory):
...
+
+    Path('pipeline.serve.yaml').write_text(f"""
+tasks:
+    - source: {'.'.join(modules)}.{function_name}
+      product: out.ipynb
+""")
+
+    runner = CliRunner()
+    result = runner.invoke(scaffold,
+                           args=['-e', 'pipeline.serve.yaml'],
```
edublancas, line 265: please parameterize this test case, one should not pass any args so ploomber uses the default pipeline.yaml

edublancas (changes requested):
pls add test cases

tests/cli/test_scaffold.py
fferegrino: cool

tests/cli/test_scaffold.py
fferegrino: cool

tests/assets/spec-with-functions-no-sources/pipeline.yaml
```diff
@@ -1,8 +1,8 @@
 tasks:
-  - source: my_tasks.raw.functions.function
+  - source: my_tasks2.raw.functions.function
     product: some_file.txt
     name: raw
   
-  - source: my_tasks.clean.functions.function
+  - source: my_tasks2.clean.functions.function
     product: another.txt
```
fferegrino, line 7: I've changed these assets so that the tests pass, we can maybe revisit it in this ticket? https://github.com/ploomber/ploomber/issues/442

edublancas: I think this is a different problem. see #449

edublancas: but this works for now, thanks!

edublancas (approved):

PR #451, opened 2021-12-26
Update .gitignore to exclude .idea/

fferegrino:
Can we exclude PyCharm's config folder just like we do with VSCode's?


edublancas:
Sure!

PR #452, opened 2021-12-27
Deprecate lazy_import

fferegrino:


src/ploomber/spec/dagspec.py
```diff
@@ -467,8 +501,8 @@ def _to_dag(self):
             if attr in self:
                 setattr(
                     dag, attr,
-                    dotted_path.DottedPath(self[attr],
-                                           lazy_load=self._lazy_import))
+                    dotted_path.DottedPath(
+                        self[attr], lazy_load=self._import_mode != 'eager'))
```
fferegrino, line 505: I am not sure `DottedPath` needs to know more than a boolean: "should I load this module now or not?" so I've preserved the lazy load logic where we don't need to explicitly skip the loading entirely.

idomic:
@fferegrino can you review the errors and make sure the CI passes so we can review this?

idomic:
Closing due to inactivity, feel free to open once you want to pick this up!

PR #463, opened 2022-01-04
Telemetry env var

edublancas:


src/ploomber/telemetry/telemetry.py
```diff
@@ -101,7 +101,16 @@ def check_uid():
...
-def check_stats_file():
+def check_stats_enabled():
+    """
+    Check if the user allows us to use telemetry. In order of precedence:
+
+    1. If PLOOMBER_STATS_ENABLED defined, check its value
+    2. Otherwise use the value in stats_enabled in the config.yaml file
+    """
+    if 'PLOOMBER_STATS_ENABLED' in os.environ:
+        return os.environ['PLOOMBER_STATS_ENABLED'].lower() == 'true'
```
idomic, line 112: We probably need to double check the original file is also set to lower() to follow the same convention

edublancas: but in YAML, there are valid booleans `false`/`true`. With env vars, they're all strings, that's why I added the lower

idomic:
Looking good

PR #464, opened 2022-01-04
disaable stats from CI

edublancas:
I disabled the stats in the CI. Needed to change some of the telemetry_tests (just minor changes). Please review for your awareness.

I also saw you're working on the stats improvements, once this PR is merged, you may want to rebase to get the new `test_telemetry.py`

### DIFF
--- a/tests/telemetry/test_telemetry.py
+++ b/tests/telemetry/test_telemetry.py
@@ -12,6 +12,26 @@ import ploomber.dag.dag as dag_module
 from conftest import _write_sample_conda_env, _prepare_files
 
 
+@pytest.fixture
+def ignore_ploomber_stats_enabled_env_var(monkeypatch):
+    """
+    GitHub Actions configuration scripts set the PLOOMBER_STATS_ENABLED
+    environment variable to prevent CI events from going to posthog, this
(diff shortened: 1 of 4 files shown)


tests/telemetry/test_telemetry.py
```diff
@@ -40,11 +60,9 @@ def test_opt_str_validation():
     assert type(exception_raised) == TypeError
 
 
-# Test class functions
-def test_check_stats_enabled():
+def test_check_stats_enabled(ignore_env_var_and_set_tmp_default_home_dir):
     stats_enabled = telemetry.check_stats_enabled()
-    assert isinstance(stats_enabled, bool)
-    assert stats_enabled
+    assert stats_enabled is True
```
idomic, line 65: What's wrong with asserting the inner value?

idomic (approved):
lgtm besides this inline comment above

tests/telemetry/test_telemetry.py
edublancas: nothing wrong but `statis_enabled is True` covers the previous two lines (is bool and is True), it's just shorter and more explicit :)

edublancas:
re: ploomber scaffold. good catch. I haven't applied it yet. will do!

PR #465, opened 2022-01-05
Stats improvements

idomic:


src/ploomber/telemetry/telemetry.py
```diff
@@ -62,6 +63,50 @@ def is_online(host='http://ploomber.io'):
         return False
 
 
+# Will output if the code is within a container
```
edublancas, line 66: can you turn these comments into docstrings?

src/ploomber/telemetry/telemetry.py
```diff
@@ -62,6 +63,50 @@ def is_online(host='http://ploomber.io'):
...
+def is_docker():
+    path = '/proc/self/cgroup'
+    return (os.path.exists('/.dockerenv')
```
edublancas, line 69: not a big deal but consider using pathlib.Path instead of os.path, it has a cleaner API and it's what we use in other modules

src/ploomber/telemetry/telemetry.py
```diff
@@ -62,6 +63,50 @@ def is_online(host='http://ploomber.io'):
...
+    if os == "Darwin":
+        return 'MacOS'
+    else:  # Windows/Linux are contained
+        return os
+
+
+# The function will tell if the code is running in a conda env
+def is_conda():
+    return os.path.exists(os.path.join(sys.prefix, 'conda-meta')) or \
+            bool(os.environ.get("CONDA_PREFIX")) or \
```
edublancas, line 86: no need to use `bool()` here `None` evaluates to `False`. Also, the other modules use parenthesis to break down large statements (instead of `\`)

src/ploomber/telemetry/telemetry.py
```diff
@@ -180,6 +239,10 @@ def log_api(action, client_time=None, total_runtime=None, metadata=None):
                             'total_runtime': total_runtime,
                             'python_version': py_version,
                             'ploomber_version': product_version,
+                            'docker_container': docker_container,
+                            'operating_system': operating_system,
+                            'environment': environment,
+                            'is_install': is_install,
```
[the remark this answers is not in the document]
edublancas: which parameter?

tests/telemetry/test_telemetry.py
```diff
@@ -12,6 +12,31 @@
 from conftest import _write_sample_conda_env, _prepare_files
 
 
+# The below fixtures are to mock the different virtual environments
+# Ref: https://stackoverflow.com/questions/51266880/detect-if-
+# python-is-running-in-a-conda-environment
```
edublancas, line 17: thanks for including the reference, i think this and the next one should go next to the function that checks the conditions, rather than the fixtures

tests/telemetry/test_telemetry.py
```diff
@@ -12,6 +12,31 @@
...
+    monkeypatch.setattr(telemetry.sys, 'base_prefix', 'base_prefix')
+
+
+# Ref: https://stackoverflow.com/questions/43878953/how-does-one-detect-if-
+# one-is-running-within-a-docker-container-within-python
+@pytest.fixture()
+def inside_docker(monkeypatch, tmp_directory):
+    mock = Mock()
+    mock.return_value = True
+    monkeypatch.setattr(telemetry.os.path, 'exists', mock)
```
edublancas, line 37: this fixture will return true for all calls to .exists(), but you only want a subset of those. check out Mocks documentation, there is a way to define an output when the function is called with certain inputs. e.g. return True if .exists is called with '/.dockerenv' but false with all other values

edublancas (changes requested):

src/ploomber/telemetry/telemetry.py
```diff
@@ -129,16 +174,26 @@ def check_stats_enabled():
             return f"Error: Can't read config file {e}"
 
 
+def check_first_time_usage():
+    config_path = os.path.join(check_dir_file_exist(CONF_DIR), 'config.yaml')
+    uid_path = os.path.join(check_dir_file_exist(CONF_DIR), 'uid.yaml')
+    return not os.path.exists(uid_path) and os.path.exists(config_path)
```
edublancas, line 180: shouldn't this be `not os.path.exists(uid_path) and not os.path.exists(config_path)`?

tests/telemetry/test_telemetry.py
idomic: Wow this was a hard one. Not necessarily worth it since it's isolated in the test.

src/ploomber/telemetry/telemetry.py
idomic: No since the call is being performed after the check for a stats config file (check the calling function)

src/ploomber/telemetry/telemetry.py
```diff
@@ -107,39 +114,41 @@ def get_env():
...
     if PLOOMBER_HOME_DIR:
-        final_location = os.path.expanduser(PLOOMBER_HOME_DIR)
+        final_location = PLOOMBER_HOME_DIR
     else:
-        final_location = os.path.expanduser(DEFAULT_HOME_DIR)
+        final_location = DEFAULT_HOME_DIR
 
     if name:
-        final_location = os.path.join(final_location, name)
+        final_location = f"{final_location}/{name}"
```
edublancas, line 128: better to use `Path(final_location, name)` instead of `/` - since it may break on windows

src/ploomber/telemetry/telemetry.py
```diff
@@ -107,39 +114,41 @@ def get_env():
...
+    p = Path(final_location)
```
edublancas, line 129: i think you need `Path(final_location).expanduser()` here

src/ploomber/telemetry/telemetry.py
```diff
@@ -107,39 +114,41 @@ def get_env():
...
-    if not os.path.exists(uid_path):  # Create - doesn't exist
+    """
+    Checks if local user id exists as a uid file, creates if not.
+    """
+    uid_path = Path(check_dir_exist(CONF_DIR), 'uid.yaml')
+    if not uid_path.exists():  # Create - doesn't exist
         uid = str(uuid.uuid4())
         try:  # Create for future runs
-            with open(uid_path, "w") as file:
+            with uid_path.open("w", encoding="utf-8") as file:
```
edublancas, line 144: can we leave the default encoding instead of hardcoding? this is system-dependent. no need to change it but for future reference you can use `Path.write_text` and `Path.read_text` instead of `Path.open`

src/ploomber/telemetry/telemetry.py
```diff
@@ -157,43 +166,51 @@ def check_stats_enabled():
         return os.environ['PLOOMBER_STATS_ENABLED'].lower() == 'true'
 
     # Check if local config exists
-    config_path = os.path.join(check_dir_file_exist(CONF_DIR), 'config.yaml')
-    if not os.path.exists(config_path):
+    config_path = Path(check_dir_exist(CONF_DIR), 'config.yaml')
+    if not config_path.exists():
         try:  # Create for future runs
-            with open(config_path, "w") as file:
+            with config_path.open("w", encoding="utf-8") as file:
```
edublancas, line 172: can we remove the encoding?

edublancas (changes requested):

edublancas:
what happens if any of the functions that check the config fail? like `get_uid`? there is a lot of read/write logic in there which may fail due to filesystem permissions or any other thing. that should not stop the user from using the tool.

So maybe have something like this:

```
def _check_uid():
    # all the logic here
     pass

def check_uid():
    try:
        return _check_uid():
    exception Exception:
        return None
```

thoughts?

src/ploomber/telemetry/telemetry.py
idomic: Sure, where I can I'll try these functions, here I figured it's better to enforce a yaml structure via the class, without open it will not work. You can take a look of this thread: https://stackoverflow.com/questions/57296168/pathlib-path-write-text-in-append-mode

PR #488, opened 2022-01-17
Stats fixes/updates

idomic:
Updated a few env detection, more exception catching.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -153,6 +153,7 @@ setup(
         'humanize',
         'tqdm',
         'posthog',
+        'distro',
         'importlib_resources;python_version<"3.7"',
         # for code normalization, parso is also needed for inferring upstream
         # dependencies in jupyter notebooks
--- a/src/ploomber/telemetry/telemetry.py
+++ b/src/ploomber/telemetry/telemetry.py
@@ -344,11 +344,13 @@ def log_api(action,
             client_time=None,
             total_runtime=None,
             dag=None,
-            metadata={}):
+            metadata=None):
     """
     This function logs through an API call, assigns parameters if missing like
     timestamp, event id and stats information.
     """
+    metadata = metadata or {}
+
     event_id = uuid.uuid4()
     if client_time is None:
         client_time = datetime.datetime.now()
--- a/src/ploomber/telemetry/telemetry.py
+++ b/src/ploomber/telemetry/telemetry.py
@@ -31,17 +31,21 @@ The data we collect is limited to:
 
 import datetime
 import http.client as httplib
+import warnings
+
 import posthog
 import yaml
 import os
 from pathlib import Path
 import sys
 import uuid
+
 from ploomber.telemetry import validate_inputs
 from ploomber import __version__
 import platform
+import distro
 
-TELEMETRY_VERSION = '0.2'
+TELEMETRY_VERSION = '0.3'
 DEFAULT_HOME_DIR = '~/.ploomber'
 CONF_DIR = "stats"
 posthog.project_api_key = 'phc_P9SpSeypyPwxrMdFn2edOOEooQioF2axppyEeDwtMSP'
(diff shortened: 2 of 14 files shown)


idomic:
I think the CI will pass, we can start reviewing

setup.py
```diff
@@ -153,6 +153,7 @@ def read(name):
         'humanize',
         'tqdm',
         'posthog',
+        'distro',
```
edublancas, line 156: I'm not fond of adding new dependencies. I actually want to remove some of them. I saw we only use this to determine the linux version. Do you think having that level of detail is important?

src/ploomber/telemetry/telemetry.py
```diff
@@ -69,11 +72,14 @@ def is_online():
...
-            and any('docker' in line
-                    for line in cgroup.read_text().splitlines()))
+    try:
+        cgroup = Path('/proc/self/cgroup')
+        docker_env = Path('/.dockerenv')
+        return (docker_env.exists() or cgroup.exists()
+                and any('docker' in line
+                        for line in cgroup.read_text().splitlines()))
+    except FileNotFoundError:
+        return False
```
edublancas, line 82: will this ever raise FileNotFoundError? `Path(...)` will always work

src/ploomber/telemetry/telemetry.py
```diff
@@ -151,15 +248,17 @@ def check_uid():
             with uid_path.open("w") as file:
                 yaml.dump({"uid": uid}, file)
             return uid
-        except FileNotFoundError as e:
-            return f"ERROR: Can't read UID file: {e}"
+        except Exception as e:
+            print(f"ERROR: Can't write UID file: {e}")
```
edublancas, line 252: please use `warnings.warn` instead of `print`

src/ploomber/telemetry/telemetry.py
```diff
@@ -151,15 +248,17 @@ def check_uid():
...
     else:  # read and return uid
         try:
             with uid_path.open("r") as file:
                 uid_dict = yaml.safe_load(file)
             return uid_dict['uid']
-        except FileNotFoundError as e:
-            return f"Error: Can't read UID file: {e}"
+        except Exception as e:
+            print(f"Error: Can't read UID file: {e}")
+            return f"NO_UID {e}"
```
edublancas, line 261: what is this value used for? may be better to return `None`

src/ploomber/telemetry/telemetry.py
```diff
@@ -179,15 +278,17 @@ def check_stats_enabled():
             with config_path.open("w") as file:
                 yaml.dump({"stats_enabled": True}, file)
             return True
-        except FileNotFoundError as e:
-            return f"ERROR: Can't read file: {e}"
+        except Exception as e:
+            print(f"ERROR: Can't write to config file: {e}")
```
edublancas, line 282: please use warnings.warn instead of print

src/ploomber/telemetry/telemetry.py
```diff
@@ -242,7 +347,28 @@ def log_api(action, client_time=None, total_runtime=None, metadata=None):
...
+
+    airflow = is_airflow()
+    if airflow:
+        metadata['airflow'] = airflow
+
+    argo = is_argo()
+    if argo:
+        metadata['argo'] = argo
+    if dag and isinstance(dag, DAG):
+        metadata['dag'] = parse_dag(dag)
```
edublancas, line 370: maybe to a `try: parse(dag)` and store `None` if it doesn't work. no need to check if it's a DAG and we cover the case where `parse_dag` fails for any reason

tests/telemetry/test_telemetry.py
```diff
@@ -333,6 +370,98 @@ def test_is_online():
...
+                'type': 'PythonCallable',
+                'upstream': {
+                    'first': 'file1.txt',
+                    'third': 'file3.parquet'
+                },
+                'products': 'file2.txt'
+            }
+        }
+    }
+    # assert res['tasks']['first'] == {
```
edublancas, line 440: please remove this commented code

edublancas (changes requested):

setup.py
idomic: Usually when bugs happens, it's easier to recreate within the exact same env (and version). I don't think it's that crucial, let's discuss the tradeoffs.

src/ploomber/telemetry/telemetry.py
idomic: It can, it's just the wrong error. Look here: https://docs.python.org/3/library/pathlib.html (look for exist) and here https://docs.python.org/3/library/exceptions.html#OSError

It's been fixed in 3.8 but in 3.7/6 it'll still raise an OSError

setup.py
idomic: I think let's keep if for now after we saw it's not heavy.

src/ploomber/telemetry/telemetry.py
idomic: Added a flag in metadata for missing uids

PR #515, opened 2022-01-31
Added version check

idomic:


src/ploomber/telemetry/telemetry.py
```diff
@@ -310,6 +311,71 @@ def check_first_time_usage():
...
+    If it's not the latest, notifies the user and saves the metadata to conf
+    Alerting every 2 days on stale versions
+    """
+    # Read conf file
+    config_path = Path(check_dir_exist(CONF_DIR), 'config.yaml')
+    try:
+        with config_path.open("r") as file:
+            conf = yaml.safe_load(file)
+    except Exception as e:
+        warnings.warn(f"Error: Can't read config file {e}")
```
edublancas, line 344: i've seen this reading conf block in the telemetry code, can you create a function and then call it here and when reading it for the telemetry?

src/ploomber/telemetry/telemetry.py
```diff
@@ -310,6 +311,71 @@ def check_first_time_usage():
...
+
+    # If the version is updated continue
+    if 'latest_ploomber_version' not in conf.keys():
+        latest = get_latest_version()
+        conf['latest_ploomber_version'] = latest
+    else:
+        latest = conf['latest_ploomber_version']
```
edublancas, line 351: why do we need to save the latest version(instead of just pinging pypi again)?

src/ploomber/telemetry/telemetry.py
```diff
@@ -310,6 +311,71 @@ def check_first_time_usage():
...
+    warnings.warn("****Please update the ploomber version, run:\n"
+                  "pip install ploomber --upgrade\n")
+
+    # Update conf
+    conf['last_version_check'] = today
+    try:  # Create for future runs
+        with config_path.open("w") as file:
+            yaml.dump(conf, file)
+    except Exception as e:
+        warnings.warn(f"ERROR: Can't write to config file: {e}")
```
edublancas, line 376: same here. I think reading and writing to the conf file should  be abstracted in a function

tests/telemetry/test_telemetry.py
```diff
@@ -458,6 +464,72 @@ def test_validate_entries(monkeypatch):
     assert res == (event_id, uid, action, client_time, elapsed_time)
 
 
+def test_get_latest_version(monkeypatch):
+    is_latest = telemetry.get_latest_version()
+    assert isinstance(is_latest, str)
+    version_index = [i for i, ltr in enumerate(is_latest) if ltr == '.']
+    assert len(version_index) == 2
```
edublancas, line 471: what are these two assert statements checking? they don't seem important, the latest one does `assert is_latest == ploomber.__version__`

tests/telemetry/test_telemetry.py
```diff
@@ -458,6 +464,72 @@ def test_validate_entries(monkeypatch):
...
+    # Mock version and the conf, check it produces the same version
+
+
+def write_to_conf_file(path, version, last_check):
+    (path).write_text("stats_enabled: True\n"
+                      f"last_version_check: {last_check}\n"
+                      f"latest_ploomber_version: {version}\n")
+
+
+def test_check_version(tmp_directory, monkeypatch):
```
edublancas, line 488: can you break this test into many small tests? It's best for readability to have each test case do one thing, you can use fixtures to avoid having redundant code

tests/telemetry/test_telemetry.py
```diff
@@ -458,6 +464,72 @@ def test_validate_entries(monkeypatch):
...
+    # (path).write_text("stats_enabled: True\n"
+    #                   f"last_version_check: 2022-01-20 10:51:41.082376\n"
+    #                   f"latest_ploomber_version: 0.14.8\n")
+    write_to_conf_file(path=path,
+                       version='0.14.8',
+                       last_check='2022-01-20 10:51:41.082376')
+
+    # Test no warning when same version encountered
+    # with warnings.catch_warnings():
+    #     warnings.filterwarnings("error")
```
edublancas, line 504: remove all commented code

edublancas (changes requested):

src/ploomber/telemetry/telemetry.py
idomic: No real need, I thought I'd reuse the version (faster) instead of keep sending those pipy http requests

tests/telemetry/test_telemetry.py
idomic: So this is commented since I couldn't mock the __version__ to a specific version for some reason, I'm not sure what's going on there

tests/telemetry/test_telemetry.py
idomic: Sure.

tests/telemetry/test_telemetry.py
idomic: So it's to test that the pipy http call returned a valid version of type string and it has the right formatting either 1 or 2 .'s

edublancas (changes requested):

edublancas:
closing this in favor of #558

PR #516, opened 2022-01-31
Contributing updates

maticortesr:
Minor details to smooth out on-boarding


edublancas:
thanks for the contribution! did you encounter any issues that were resolved by adding conda-forge?

maticortesr:
Yes, on my Fedora machine the `invoke setup` would get stuck on an infinite wait, having to kill the terminal. The couple error messages before getting stuck where very useless as well so worth adding a note

edublancas:
oh, great to know!  thanks for adding this!

PR #517, opened 2022-02-01
Improved error message when pipeline.yaml does not exist

### LINKED ISSUE #509: improve error message when pipeline.yaml does not exist
If a user runs any ploomber command and passes `-e pipeline.yaml` in a directory that doesn't have such a file, the error isn't very clear:

```sh
ploomber status -e pipeline.yaml
```

```pytb
  File "/Users/Edu/dev/ploomber/src/ploomber/cli/io.py", line 34, in wrapper
    fn(**kwargs)
  File "/Users/Edu/dev/ploomber/src/ploomber/cli/status.py", line 15, in main
    dag, args = parser.load_from_entry_point_arg()
  File "/Users/Edu/dev/ploomber/src/ploomber/cli/parsers.py", line 213, in load_from_entry_point_arg
    entry_point = EntryPoint(self.parse_entry_point_value())
  File "/Users/Edu/dev/ploomber/src/ploomber/entrypoint.py", line 19, in __init__
    self.type = find_entry_point_type(value)
  File "/Users/Edu/dev/ploomber/src/ploomber/entrypoint.py", line 65, in find_entry_point_type
    raise ValueError(
ValueError: Could not determine the entry point type from value: 'pipeline.yaml'. Expected an existing file with extension .yaml or .yml, existing directory, glob-like pattern (i.e., *.py) or dotted path (i.e., module.sub_module.factory_function). Verify your input.
```

It'd be better to check if the argument "looks like" a path to a yaml file, and if so, say that the path doesn't exist. If it doesn't look like a path to a yaml, print the default error message.

aadityasinha-dotcom:
Improved error message when pipeline.yaml does not exist using `pathlib.Path`.

Resolves #509

### DIFF
--- a/src/ploomber/entrypoint.py
+++ b/src/ploomber/entrypoint.py
@@ -63,7 +63,7 @@ def find_entry_point_type(entry_point):
         return type_
     else:
         if Path(entry_point).suffix == '.yaml':
-            raise Exception('Could not determine the entry point type from value: ' 
+            raise ValueError('Could not determine the entry point type from value: ' 
             f'{entry_point!r}. The file does not exist.')
         else:
             raise ValueError(
--- a/src/ploomber/entrypoint.py
+++ b/src/ploomber/entrypoint.py
@@ -62,13 +62,16 @@ def find_entry_point_type(entry_point):
     if type_:
         return type_
     else:
-        raise ValueError(
-            'Could not determine the entry point type from value: '
-            f'{entry_point!r}. Expected '
-            'an existing file with extension .yaml or .yml, '
-            'existing directory, glob-like pattern '
-            '(i.e., *.py) or dotted path '
-            '(i.e., module.sub_module.factory_function). Verify your input.')
+        if Path('my_file.yaml').suffix == '.yaml':
+            raise Exception("Path does'nt exist")
+        else:
+            raise ValueError(
+                'Could not determine the entry point type from value: '
+                f'{entry_point!r}. Expected '
+                'an existing file with extension .yaml or .yml, '
+                'existing directory, glob-like pattern '
+                '(i.e., *.py) or dotted path '
+                '(i.e., module.sub_module.factory_function). Verify your input.')
 
 
 def try_to_find_entry_point_type(entry_point):
--- a/src/ploomber/entrypoint.py
+++ b/src/ploomber/entrypoint.py
@@ -62,8 +62,9 @@ def find_entry_point_type(entry_point):
     if type_:
         return type_
     else:
-        if Path('my_file.yaml').suffix == '.yaml':
-            raise Exception("Path does'nt exist")
+        if Path(entry_point).suffix == '.yaml':
+            raise Exception('Could not determine the entry point type from value: ' 
+            f'{entry_point!r}. The file does not exist.')
         else:
             raise ValueError(
                 'Could not determine the entry point type from value: '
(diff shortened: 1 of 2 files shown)


src/ploomber/entrypoint.py
```diff
@@ -62,13 +62,16 @@ def find_entry_point_type(entry_point):
...
         return type_
     else:
-        raise ValueError(
-            'Could not determine the entry point type from value: '
-            f'{entry_point!r}. Expected '
-            'an existing file with extension .yaml or .yml, '
-            'existing directory, glob-like pattern '
-            '(i.e., *.py) or dotted path '
-            '(i.e., module.sub_module.factory_function). Verify your input.')
+        if Path('my_file.yaml').suffix == '.yaml':
```
edublancas, line 65: this should be `Path(entry_point).suffix`

edublancas: also, add a test here: https://github.com/ploomber/ploomber/blob/master/tests/test_entrypoint.py

src/ploomber/entrypoint.py
```diff
@@ -62,13 +62,16 @@ def find_entry_point_type(entry_point):
...
+            raise Exception("Path does'nt exist")
```
edublancas, line 66: please add a more meaningful error message like: "Could not determine entry point from value ...: the file does not exist"

edublancas (changes requested):

src/ploomber/entrypoint.py
aadityasinha-dotcom: Sorry, forgot to change that.

aadityasinha-dotcom:
@edublancas Can you review those changes?

src/ploomber/entrypoint.py
```diff
@@ -62,13 +62,17 @@ def find_entry_point_type(entry_point):
...
     else:
-        raise ValueError(
-            'Could not determine the entry point type from value: '
-            f'{entry_point!r}. Expected '
-            'an existing file with extension .yaml or .yml, '
-            'existing directory, glob-like pattern '
-            '(i.e., *.py) or dotted path '
-            '(i.e., module.sub_module.factory_function). Verify your input.')
+        if Path(entry_point).suffix == '.yaml':
+            raise Exception('Could not determine the entry point type from value: '
```
edublancas, line 66: this looks a lot better! :) please change this to ValueError

tests/test_entrypoint.py
```diff
@@ -33,3 +33,6 @@ def test_dotted_path_that_ends_with_yaml():
 
 def test_try_to_find_entry_point_type_with_none():
     assert try_to_find_entry_point_type(None) is None
+
+def test_find_entry_point_type_with_none():
```
edublancas, line 37: the test cases should be representative of what we wanna test, so try passing some value like 'some.value' and check that you get certain error type and error message, then pass something like 'file.yaml' and test that you get the other errors message 

here's an example:

```py
def test_recursion_depth():
    with pytest.raises(RuntimeError) as excinfo:

        def f():
            f()

        f()
    assert "maximum recursion" in str(excinfo.value)
```

source: https://docs.pytest.org/en/6.2.x/assert.html#assertions-about-expected-exceptions

edublancas (changes requested):

src/ploomber/entrypoint.py
aadityasinha-dotcom: OK

tests/test_entrypoint.py
aadityasinha-dotcom: Something like this?

```
def test_find_entry_point_type_with_none():
    with pytest.raises(ValueError) as excinfo:
        find_entry_point_type('some.yaml')

    assert 'File does not exist' in str(excinfo.value)
```

edublancas:
thanks for your contribution!

PR #544, opened 2022-02-05
Log exception

### LINKED ISSUE #444: Refactor telemetry API
(Note: let's work on this after #441 is merged)

There is a common pattern in all telemetry calls:

```py
start = datetime.now()

# do stuff

now = datetime.now()

log(elapsed=now - start, dict(param=42))
```

It'd be clear to capture this logic in a context manager:

```py

with log_api(...) as logger:
    # do stuff
    ...
    
   logger.params(param=42)

```

edublancas:
closes #444 

I added a decorator that will allow us to decouple the ploomber logic from the logging logic, plus, it will allow us to capture any exception so we have more visibility

Note: I haven't started using the decorator yet, this is just the implementation


idomic (approved):
lgtm, does every error has a type? I do remember a few of those just having a generic type

edublancas:
not all of them, I'll update the ones that don't have one. Also, what happens if we log `None` to posthog? does it translate it correctly as `null`?

PR #554, opened 2022-02-07
Added auto completion with click to cli

jramirez857:
I was not sure if this feature requires tests. If so, please point me in the right direction on where tests for this feature should be added. 

One thing to note here is that I had to add `conda activate ploomber`  and  `eval "$(_PLOOMBER_COMPLETE=zsh_source ploomber)"` to my ~/.zshrc to get autocomplete to work. It does not look like this is part of the ploomber setup documentation currently so this should be added. 

Please let me know if I should update the README.md and CONTRIBUTING.md to include these steps.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ projects-tmp/
 *.parquet
 examples/pipeline/results/
 ploomber-venv
-ploomb_env
 
 doc/**/*.ipynb
 .pytest_cache/
--- a/doc/cookbook/db-config.rst
+++ b/doc/cookbook/db-config.rst
@@ -151,9 +151,11 @@ Configuring the client in ``pipeline.yaml``
 Check out the :doc:`../get-started/sql-pipeline` to learn how to configure
 the database client in your ``pipeline.yaml`` file. 
 
-SQL Examples
----------------
+Examples
+--------
+
 To see some examples using SQL connections, see this:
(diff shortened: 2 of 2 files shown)


edublancas:
thanks for your contribution! we're a bit swamped this week but I'll review this in the next few days.

jramirez857:
Thank you for the update, @edublancas! No worries.

edublancas:
Hi @jramirez857, thanks for your patience!

I did a quick test and realized that the `@click.argument("name", shell_complete=complete_env_vars)` part is overriding the autocompletion script. Example:

```sh
$ ploomber examples ex
```

Pressing tab recommends again "examples"

```sh
$ ploomber examples examples
```

tbh, I didn't understand why at the beginning but now I see why. In our case, we don't need the changes to `cli.py` since the vanilla script works for us. Can you revert the changes to `cli.py`?

Hence, the only change needed for this PR would be to document what [click's documentation](https://click.palletsprojects.com/en/8.0.x/shell-completion/#enabling-completion) suggests:

```
Add this to ~/.bashrc:
eval "$(_FOO_BAR_COMPLETE=bash_source foo-bar)"
```

for the three supported shells.

Please add it to the bottom of this document: https://github.com/ploomber/ploomber/blob/master/doc/user-guide/cli.rst

Thanks!

PR #558, opened 2022-02-07
Version update - New branch

idomic:


edublancas:
is this ready for review?

doc/community/user-stats.rst
```diff
@@ -12,3 +12,11 @@ The data we collect is limited to:
 - A generated UUID, randomized when the initial install takes place, no personal or any identifiable information.
 - Environment variables: The OS architecture Ploomber is used in (Python version etc.)
 - Information about the different product phases: installation, API calls and errors.
+
+Version updates
+***************
```
edublancas, line 17: this should be level-two header: 

```
something
-----------
```

`***` is level-3 and won't be displayed in the right navigation bar

src/ploomber/telemetry/telemetry.py
```diff
@@ -311,6 +296,84 @@ def check_first_time_usage():
...
+    # First time the version is checked
+    if 'version_check_enabled' in conf.keys():
+
+        # Check if we already notified in the last 2 days
+        last_message = conf['version_check_enabled']
+        diff = (today - last_message).days
+        if diff < 2:
+            return
+
+    click.echo("****Please update the ploomber version, run:\n"
```
edublancas, line 369: this `***` at the beginning look odd 😂 maybe remove them. also, click has a `secho` function that allows you to change the font color, i think it'd be good to show it in yellow to draw more attention to it

tests/telemetry/test_telemetry.py
```diff
@@ -464,6 +467,89 @@ def test_validate_entries(monkeypatch):
...
+    mock_version = Mock()
+    mock_version.return_value = '0.14.8'
+    monkeypatch.setattr(telemetry, 'get_latest_version', mock_version)
+
+    write_to_conf_file(
+        tmp_directory=tmp_directory,
+        monkeypatch=monkeypatch,
+        last_check='2022-01-20 10:51:41.082376')  # version='0.14.8',
+
+    # Test no warning when same version encountered
```
edublancas, line 506: remove commented code

edublancas (changes requested):

tests/telemetry/test_telemetry.py
edublancas: wait, what is this testing?

idomic: I wanted to test no echo was printed when it's the same version, for some reason the patch I tried on the ploomber.__version__ didn't work, not sure why.

tests/telemetry/test_telemetry.py
```diff
@@ -464,6 +465,89 @@ def test_validate_entries(monkeypatch):
...
+    stats.mkdir()
+    path = stats / 'config.yaml'
+    monkeypatch.setattr(telemetry, 'DEFAULT_HOME_DIR', '.')
+    (path).write_text("stats_enabled: True\n"
+                      f"version_check_enabled: {last_check}\n")
+
+
+def test_version_skips_when_updated(tmp_directory, capsys, monkeypatch):
+    # Path conf file
+    monkeypatch.setattr(telemetry, '__version__', '0.14.8')
```
edublancas, line 494: FYI @idomic : i fixed this test, instead of patching `ploomber.__version__` you need to patch the local copy, `telemetry.__version__`

edublancas:
check errors, the build is failing

edublancas:
tests pass. is this ready for review?

idomic:
Give me today to run some more tests on it :)

edublancas:
sure! I quickly skimmed through the code: can you add a test for the timeout? I see the `timeout=1` there but it'd best to explicitly test that the `get_version()` function actually times out after 1 second

idomic:
@edublancas ready for review, I've decoupled the files and added the timeouts check

src/ploomber/telemetry/telemetry.py
```diff
@@ -311,6 +296,93 @@ def check_first_time_usage():
...
+    try:
+        with conf_path.open("r") as file:
+            conf = yaml.safe_load(file)
+            return conf
+    except Exception as e:
+        warnings.warn(f"Error: Can't read config file {e}")
+        return {}
+
+
+def write_conf_file(conf_path, to_write, error=None):
```
edublancas, line 327: I think this is gonna break. because when you update the timestamp and call this function, you overwrite the file so if it has a user id, it'll be deleted. right? if so, i think this should read the current contents, mix them with the updates and write again:

```python
# something like this
to_store = {**current_file_contents, **overwritten_values}
```

please update this and add a test to ensure that calling this does not overwrite the existing content

edublancas (changes requested):

edublancas (approved):

Issue #564, opened 2022-02-08
update documents to reflect new default format


edublancas:
initially, we were using the `py:light` format as the default one for scripts:

```python
# +
x = 1

# +
y = 2
```

However, we're now migrating to the percent format to be the default, since VSCode, PyCharm, and others support it:

```python
# %%
x = 1

# %%
y = 2
```

Some sections in our documentation need updates:

https://docs.ploomber.io/en/latest/user-guide/jupyter.html
https://docs.ploomber.io/en/latest/get-started/basic-concepts.html

I found these two but there may be others.

idomic:
@hypefi wanna take a stab at it?

PR #578, opened 2022-02-10
update documents to reflect new default format #564

hypefi:
Hello, I replaced most # + with # %%, but I have let some instances of # + where both are proposed.


edublancas:
thanks a lot!

PR #588, opened 2022-02-11
Basic env check

idomic:
This relates to ploomber/soopervisor#67


src/ploomber/util/util.py
```diff
@@ -82,6 +82,17 @@ def _make_requires_error_message(missing_pkgs, fn_name, extra_msg):
...
 
+def check_mixed_envs():
+    try:
+        from pip._internal.operations import freeze
+    except ImportError:
+        from pip.operations import freeze
+    reqs = freeze.freeze()
+    for dep in reqs:
+        if ' @ file://' in dep:
+            raise CondaPipMixedEnvError(dep)
```
edublancas, line 93: maybe grab all the offending lines first so we show them all of them in the error message?

tests/util/test_default.py
```diff
@@ -521,3 +524,15 @@ def test_entry_point_relative_error_if_doesnt_exist(tmp_directory):
...
 def test_extract_name(arg, expected):
     assert default.extract_name(arg) == expected
+
+
+def test_mixed_envs(tmp_directory):
+    # Set a conda env with an install
+    _write_sample_conda_env()
+    process = subprocess.Popen(
+        "conda install pygraphviz -c conda-forge -y".split(),
+        stdout=subprocess.PIPE)
```
edublancas, line 534: this line will take a ~1min to run, it's better to just fake the file and add the `pypgraphviz @ file:` thing

src/ploomber/util/util.py
```diff
@@ -82,6 +82,20 @@ def _make_requires_error_message(missing_pkgs, fn_name, extra_msg):
     return error_msg
 
 
+def check_mixed_envs():
+    try:
+        from pip._internal.operations import freeze
+    except ImportError:
+        from pip.operations import freeze
```
edublancas, line 89: let's make this function take the output of `pip freeze` as input (the string output). because when running `ploomber install` the user may create a new environment, hence we want to check the freeze in that env, not in the current one.

```python
def check_mixed_envs(text):
    ...
```

edublancas (changes requested):

PR #608, opened 2022-02-20
Cloud pipeline monitoring

idomic:


.github/workflows/ci-mac.yml
```diff
@@ -82,6 +82,7 @@ jobs:
     - name: Run tests
       env:
         PLOOMBER_STATS_ENABLED: false
+        CI_CLOUD_API_KEY: ${{ secrets.CI_CLOUD_API_KEY }}
```
edublancas, line 85: I think this should have a more specific name, maybe `PLOOMBER_CLOUD_API_KEY`? and looks like this is also missing in the cloud docs

src/ploomber/cli/ploomber_io.py
```diff
@@ -62,7 +62,7 @@ def wrapper(catch_exception=True, **kwargs):
     return wrapper
 
 
-# FIXME: capture only certain types of exceptions. If it's something we dind't
+# FIXME: capture only certain types of exceptions. If it's something we didn't
```
edublancas, line 65: why was this renamed from io -> ploomber_io?

edublancas (changes requested):
just a few comments

src/ploomber/cli/ploomber_io.py
edublancas: this is causing some merge conflicts so if there's isn't a strong reason to rename i'd say not to. but if we want to rename, it should be something that doesn't have "ploomber" as the module name, that's a bit weird

doc/cloud/api-key.rst
```diff
@@ -0,0 +1,52 @@
+
+Install the Cloud branch
+========================
+To get this private beta functionality you'll have to install the private beta branch.
+To do so you can run this command:
+
+.. code-block:: console
+
+    pip install git+https://github.com/ploomber/ploomber@cloud
```
edublancas, line 9: pls remove this, no longer needed

doc/cloud/api-key.rst
```diff
@@ -0,0 +1,52 @@
+
+Install the Cloud branch
```
edublancas, line 2: let's mention this is a beta feature and the API may change in the future

src/ploomber/cli/ploomber_io.py
idomic: This was a pycharm bug if I recall correctly. We can keep it as is to avoid the conflicts

.github/workflows/ci-mac.yml
idomic: This is the key that gets from git secrets to run the CI why does it needs to be in the docs?

edublancas: because I see value in having this for end-users as well, give them the option to use and env variable to set the key

src/ploomber/cli/ploomber_io.py
edublancas: what error? renaming is going to cause conflicts as well. there are other branches that have updated this module

.github/workflows/ci-mac.yml
idomic: I hear you, we can change the naming here, having an env var is another feature regardless to this testing var.

src/ploomber/cli/ploomber_io.py
edublancas: oh and I just remembered that the `io` module is part of the public API, if we change this, we'll break the code of some users

Issue #429, opened 2021-11-30
Warn on large files when using resources_


edublancas:
Users can tell Ploomber to track changes to configuration files via [resources_](https://ploomber.readthedocs.io/en/latest/api/spec.html?highlight=resources_#tasks-params-resources), however, to track changes, we compute a file hash which may take too long if the file is large.

We should show a warning if this happens, `resources_` should not be used with large files.

edublancas:
the fix involves checking the file size of tasks passed in the `resources_` key (see link in the original comment for an example), to check files, use this: https://docs.python.org/3/library/stat.html

if any file is above some threshold (say >1MB), then we should [the warnings module](https://docs.python.org/3/library/warnings.html) to alert the user that the file is too big

Vinay26k:
Hi, please assign this issue to me. I will give it a try.

edublancas:
sure!

edublancas:
closed by #609

PR #609, opened 2022-02-21
fix: warn large resource issue#429

Vinay26k:
PR for Issue #429

changes in the PR:

- implementation of `_check_file_size(path)` warns user about file > 1MB
- function call before calculating hashing of resource file
-----
**pipeline.yaml**
```yaml
tasks:
  - source: tasks.raw.get
    product: products/raw/get.csv
    params:
      resources_:
        file: conf.yaml
    .....
```
**conf.yaml**: dummy file with ~9MB of data
**script used to test**:
```python
def get(resources_, product):
    print(" -------------- Debug --------------")
    print(open(resources_['file']).read()[:10])
    ....
```
**output**:
`File too large. Resource <DIR-PATH>/conf.yaml[9.98MB] > 1MB`

### DIFF
--- a/src/ploomber/products/_resources.py
+++ b/src/ploomber/products/_resources.py
@@ -2,6 +2,8 @@ from copy import deepcopy
 import hashlib
 from pathlib import Path
 from collections.abc import Mapping
+import os
+import warnings
 
 _KEY = 'resources_'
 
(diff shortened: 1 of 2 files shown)


edublancas:
looks good! can you add a test? please add it to the `test_resources.py` file

you can check for warnings using this: https://docs.pytest.org/en/6.2.x/warnings.html

and to simulate a big file, you can patch the `os.stat` call:

```python
from ploomber.products import _resources

def test_process_resources_warns_on_large_file(monkeypatch):
    # this will cause the os.stat call in the _resources module to return 2E+6
    monkeypatch(_resources.os, 'stat', lambda _: 2E+6)
    # check warning is displayed
```

edublancas (approved):

edublancas:
thanks a lot for your contribution!

PR #665, opened 2022-03-19
Updated celll to cell. Fixes #664

### LINKED ISSUE #664: Celll should be replaced by cell
[Here](https://github.com/ploomber/ploomber/blob/9e7ed3ffa9c8827026f0454c2c6196e555b4db1d/src/ploomber/cli/nb.py#L323) , celll should be replaced by cell. 

Can I solve this issue?

rishav-karanjit:
This commit closes #664


edublancas:
good catch, thanks!

Issue #540, opened 2022-02-04
replace dependency on pygraphviz


edublancas:
`ploomber plot` depends on pygraphviz and graphviz, the latter is not pip-installable which makes setup difficult. we should find a way for users to easily create a dag plot, ideally with no extra dependencies

edublancas:
I did some research on this. According to this [article](https://wimyedema.medium.com/drawing-dags-5cadcb452115), the best non-pygraphviz option is [d3-dag](https://erikbrinkman.github.io/d3-dag/index.html). That could work.

I think we should also implement some text-based representation that does not depend on any external libraries.

idomic:
@Vinay26k this is an easy and an important one you can work on.

Vinay26k:
Sure, assign it to me. I will give it a try.

edublancas:
this involves a bit of research and then the implementation. please share your findings so we are on the same page and we can provide some guidance. thanks a lot for your hard work!

idomic:
Until it's being fixed, if anyone is hitting issues on python < 3.8 this is the way to fix it:
sudo apt-get -qq update -y
sudo apt-get -qq install -y graphviz graphviz-dev
pip -q install graphviz 'pygraphviz<1.8'

idomic:
@Vinay26k any progress here?

Vinay26k:
Hi @idomic, yes started exploring on d3-dag and attached is the output html file rendering generated dynamically through python code. [**WIP**]
![image](https://user-images.githubusercontent.com/22490030/156888858-41e33337-1875-4667-bba7-f6354115e800.png)

idomic:
@Vinay26k This is great!
Is there a way to print the products of the tasks as well? 
Does the colors on the paths mean anything?
Also the color are still indicating the execution status right?

edublancas:
@Vinay26k has been doing great progress. so here are some pointers to incorporate his code into Ploomber:

we'll bundle an HTML template that will be used to generate the DAG, we can put this in the [resources directory](https://github.com/ploomber/ploomber/tree/master/src/ploomber/resources)

I'm thinking that the code to generate the plot should be in a separate module, say `src/ploomber/dag/plot.py`:

```python
# plot.py
def with_d3(dag):
    """Returns an HTML string with the plot, given the DAG
    """
    pass
```

@Vinay26k : I'm guessing you are thinking of using jinja to take the template and generate the output HTML?

[...]

PR #666, opened 2022-03-19
replace dependency on pygraphviz - d3 dag visualization

Vinay26k:
Issue: #540 
**Changes**:

1. cli/plot.py contains new backend argument to handle user custom input
2. dag/dag.py contains function `plot(... backend,....)` and based on backed available logic uses either pygraphviz or d3 html template
3. dag/plot.py added d3 dag related functions for parsing and generating output file
4. address `dag_template.html` file as new resource for generating d3 dag output file

**Note**:
 - Code refactoring is in progress

**References** :
- [d3-dag package repository](https://github.com/erikbrinkman/d3-dag)


edublancas (changes requested):
Looking good! I just added a small comment. Please add a few tests to check the different scenarios, check out the existing tests that we have (both when invoked from the CLI and when calling `dag.plot`

edublancas:
a few things I found:

the arrows are reversed (it should be get -> features, get -> join)

image

I tried with a larger dag and the labels don't appear:

image

just a heads up, I see this is still WIP :)

idomic:
@Vinay26k any updates here?

edublancas:
hey @Vinay26k , thanks for your work, we'll take it from here!

PR #667, opened 2022-03-20
Improvements to the {{git}} placeholder feature

edublancas:
This PR improves the `{{git}}` placeholder feature:

* Adds `{{git_hash}}` which is slightly different in behavior
* Outputs better error messages (e.g. git not installed, not in a valid repo)
* No longer requires an `env.yaml`: the placeholder can be directly used in `pipeline.yaml`

For a usage example, see: https://github.com/ploomber/projects/blob/versioning/guides/versioning/README.ipynb


idomic (approved):
LGTM, just make sure the CI is passing

Issue #648, opened 2022-03-12
Setting task class default init parameters


edublancas:
(note that this is related to #513)

In some cases, we want to customize how a task is initialized, for example, if we want to hide the code in an output HTML report, we can create a task like this:

```yaml
tasks:
  - source: fit.py
    product:
        nb: output/nb.html
        model: output/model.pickle
    
    # hide code
    nbconvert_export_kwargs:
      exclude_input: True
```

However, if we have multiple tasks, and want to hide the code in all the outputs, we need to pass the initialization parameters to all of them, which is too verbose. Alternatively, we can provide a way to pass default initialization parameters:

```yaml
init_defaults:
    # passes to all tasks
    Task:
        params: {a: 1}
    # passes to all notebook tasks
    NotebookRunner:
        nbconvert_export_kwargs:
            exclude_input: True
```

Notes

* this conflicts with `clients`, perhaps throw an error if clients appears here and tell the user to pass them in the clients section
* do not allow certain arguments. For example all __init__ methods take a dag argument but that should be allowed here
* this requires knowledge of the names of the underlying classes. is there any way to make this simpler?

idomic:
@jramirez857 want to take a look?

jramirez857:
> @jramirez857 want to take a look?

Yes, I'll try this out

jramirez857:
Hi, @edublancas @idomic . I have a couple questions

1. Am I correct in understanding the yaml file would look something like this?

```YAML
init_defaults:
    # passes to all tasks
    Task:
        params: {a: 1}
    # passes to all notebook tasks
    NotebookRunner:
        nbconvert_export_kwargs:
            exclude_input: True
        static_analysis: regular
    SQLUpload:
        chunksize : 20000
tasks:
  - source: fit.py
    product:
        nb: output/nb.html
        model: output/model.pickle
    
```

2. If the init_defaults section is new to the YAML, which other parts of the codebase would this affect (e.g. DAG or loader)? 
3. Would this only encompass parameters in tasks.py, sql.py and aws.py classes? 

I appreciate your help as I am familiarizing myself with the codebase!

edublancas:
Sure, so the changes need to go in `dagspec.py` and `taskspec.py`, the former deals with the `pipeline.yaml` spec, while the latter with each entry in the `tasks` section.

Now that I think about it `task_defaults` sounds like a better name (instead of `init_defaults`). What do you think?

[...]

PR #668, opened 2022-03-20
Task default parameters

jramirez857:
I still have to do the last part of https://github.com/ploomber/ploomber/issues/648#issuecomment-1073143121 which is to take the class name with the sub-sections in task_defaults and modify the call to the class_ constructor in task_spec but I am opening this up for initial comments.


jramirez857:
Hi @edublancas I will be cleaning this up and fixing the CI failures with flake8. Have you considered pre-commit for keeping consistent formatting with flake8? That aside, I could use some help with the last part of https://github.com/ploomber/ploomber/issues/648#issuecomment-1073143121 about passing the `task_defaults` in to `_init_task`.  I have tried modifying `TaskSpec.to_task` to initialize the task on line 323 with the task_defaults. 
```PYTHON
return _init_task(data=data,
                             meta=self.meta,
                              task_defaults=self.task_defaults,
                              project_root=self.project_root,
                              lazy_import=self.lazy_import,
                              dag=dag), upstream
```
I am also changing the function signature of `_init_task` to take in the data which I am assuming is a dictionary. I am then passing the `task_defaults` into the class by using the `class_` as a key to `task_defaults` as shown below.
```PYTHON
try:
        task = class_(task_defaults[class_],
                    source=source,
                      product=product,
                      name=name,
                      dag=dag,
                      **task_dict)
```
However, this is causing my tests to fail with DAGSpecInitializationError. Could you help me with understanding how I should be initializing the task?

edublancas:
re pre-commit: yes, we have an optional [pre-commit hook](https://github.com/ploomber/ploomber/blob/master/.githooks/pre-push), but it only checks flake8, it doesn't apply yapf. if you're up for the challenge, PRs are welcome :)

re initialization, so `task_dict` is what the user passed to the specific task and `task_defaults[class_]` is what they specified as defaults, you can combine both with something like this:

```python
init_params = {**task_defaults[class_], **task_dict}

# then intiialize 
task = class_(source=source,
                      product=product,
                      name=name,
                      dag=dag,
                      **init_params)
```

the first line is combining the two dictionaries into one, if there are overlapping keys, the final dictionary will give preference to the ones in `task_dict`, then you pass this combined dictionary to `class_` this should give you what you want

bear in mind that we have to do some here: we must check that `source`, `product`, `name` and `dag` do not appear in `task_defaults[class_]`, we can add this logic later, but keep that in mind.

src/ploomber/spec/dagspec.py
```diff
@@ -565,7 +566,7 @@ def from_directory(cls, path_to_dir):
             The directory to use.  Looks for scripts
             (``.py``, ``.R`` or ``.ipynb``) in the directory and interprets
             them as task sources, file names are assigned as task names
-            (without extension). The spec is generated with the default values
+            (without extension). The spec is generated with the default valueq2 A2AAAAAAs
```
edublancas, line 569: looks like some debugging happened here haha

tests/spec/test_taskspec.py
```diff
@@ -169,8 +169,15 @@ def test_initialization(spec, expected, tmp_sample_tasks, tmp_imports):
         'extract_product': False,
         'extract_upstream': True
     })
+    task_defaults = {
```
edublancas, line 172: glad to see you're adding some tests! a general recommendation: please keep existing tests as they are since we want each test case to only test one thing. make this argument optional so you don't need to modify these tests.

then, create new tests to check the feature you're working on

edublancas (changes requested):

src/ploomber/spec/dagspec.py
jramirez857: Haha my keyboard has a mind of its own sometimes

tests/spec/test_taskspec.py
jramirez857: Will do. I modified them since I was not totally sure that it was an optional feature. Thanks for clarifying

edublancas: thanks!

idomic:
@jramirez857 any updates there? :)

idomic:
No worries, thanks for the update!

jramirez857:
Hi @edublancas, I have updated `taskspec.py` and `dagspec.py` based on the feedback you provided. I had intended to add tests for this initially but upon a second look at existing tests in `test_taskspec.py` I do not see any type of tests of similar type for testing an optional parameter. What are your thoughts? I can now also work on checking that `source`, `product`, `name` and `dag` do not appear in `task_defaults[class_]`. How would you recommend this be handled if they do appear? Should an error be raised?

jramirez857:
@idomic please let me know if there are any other issues I can help with in the meantime 😄

src/ploomber/spec/taskspec.py
```diff
@@ -200,11 +200,13 @@ def __init__(self,
                  meta,
                  project_root,
                  lazy_import=False,
-                 reload=False):
+                 reload=False,
+                 task_defaults=""):
```
edublancas, line 204: it'd be better to have `task_defaults=None` here

src/ploomber/spec/taskspec.py
```diff
@@ -200,11 +200,13 @@ def __init__(self,
...
         self.data = deepcopy(data)
         self.meta = deepcopy(meta)
         self.project_root = project_root
         self.lazy_import = lazy_import
+        self.task_defaults = task_defaults
```
edublancas, line 209: and here `self.task_defaults = task_defaults or {}` so if passed None, then it initializes an empty dictionary

src/ploomber/spec/taskspec.py
```diff
@@ -357,7 +359,7 @@ def __repr__(self):
         return '{}({!r})'.format(type(self).__name__, self.data)
 
 
-def _init_task(data, meta, project_root, lazy_import, dag):
+def _init_task(data, meta, project_root, lazy_import, dag, task_defaults=""):
```
edublancas, line 362: same here `task_defaults=None`

src/ploomber/spec/taskspec.py
```diff
@@ -401,12 +403,17 @@ def _init_task(data, meta, project_root, lazy_import, dag):
         task_dict['params'] = resolve_resources(task_dict['params'],
                                                 relative_to=project_root)
 
+    if task_defaults != "":
```
edublancas, line 406: the condition here should be to check if we have defaults for the given class:

```
if class_.__name__ in task_defaults:
   ...
```

you can even remove the if statement:


```
init_params = {**tak_defaults.get(class_.__name__ , {}), **task_dict}
```

edublancas (changes requested):

edublancas:
> I can now also work on checking that `source`, `product`, `name` and `dag` do not appear in `task_defaults[class_]`. How would you recommend this be handled if they do appear? Should an error be raised?

yeah, an error like "task defaults should not contain the following keys:  A, B, C". I think we should add this validation earlier in the code when we read the user's spec (so in DAGSpec.py). I usually validate these things with sets: https://docs.python.org/3/tutorial/datastructures.html#sets

```python
dict_to_test = dict(a=1, b=2, z=3)
allowed_keys = {'a', 'b'}

unexpected = set(dict_to_test) - allowed_keys

if unexpected:
   raise ValueError(f'got unexpected: {unexpected}')

```

> `` I do not see any type of tests of similar type for testing an optional parameter.

can you elaborate? what's the optional parameter?

src/ploomber/spec/taskspec.py
jramirez857: Done!

edublancas:
hi @jramirez857, thanks for the patience, the changes look good. can you fix the tests? I saw the logs and noticed this error:

```
 >       init_params = {**task_defaults[class_], **task_dict}
E       TypeError: 'NoneType' object is not subscriptable
```

jramirez857:
Hi Eduardo, thanks for your patience with this pull request. What should we take in as a default value to the ` _init_task` method if no `task_defaults` are passed in. I was using a default value of `None` in the function signature which is why this error is thrown. I think that is what I meant in my previous question about the optional parameter. I was referring to making `task_default` optional 😅

jramirez857:
I have also reran these tests today with the lines below and I weirdly do not see any of the tests failing under tests/spec/test_taskspec. Do you know which specific test you were seeing failing? I have checked ploombers CI output but I see unrelated failures in the 3.6 tests.
```
def _init_task(data, meta, project_root, lazy_import, dag, task_defaults=None):
...
 >       init_params = {**task_defaults[class_], **task_dict}
E       TypeError: 'NoneType' object is not subscriptable
```

edublancas:
the default `None` is fine, what we need to cover is this error:

```
 >       init_params = {**task_defaults[class_], **task_dict}
E       TypeError: 'NoneType' object is not subscriptable
```

so here task_defaults is empty, which means the user doesn't have a task_defaults section, so you need to cover that scenario and make `init_params = task_dict`, another scenario is that task_defaults is a dictionary, but does not contain the `class_` key, in that case, again, we need to make `init_params = task_dict`

I'm not following the CI question: I looked at the CI and saw the `TypeError: 'NoneType' object is not subscriptable` error a few times, but there might be other tests with different errors, so those should be addressed as well.

jramirez857:
@edublancas I am now able to see the `TypeError: 'NoneType' object is not subscriptable` failure in the local tests after switching from codespaces to a local clone and have handled the case where there is no task_defaults section. In this case I check if a dict was passed in and set `init_params = task_dict`. I can continue fixing any failing tests from the CI/CD and add tests for the no defaults case as well as task defaults passed in cases.

edublancas:
awesome, thanks for the update! I see that some of the `test_cloud.py` fail, this is because of some misconfiguration on our end, so there's no need for you to debug those. The only remaining thing to close this PR is to add some tests for this new feature.

Essentially we want to input something like:

```
task_defaults:
   NotebookRunner:
       check_if_kernel_installed: false

tasks:
  - source: script.py
    product: output/report.html
```

Then load that `pipeline.yaml`, convert it into a DAG, and check that the task has indeed the default parameters.

Check out `test_dagspec.py` for tests with similar structure

jramirez857:
Hello, here is the task_default_parameters test I am debugging for context. 
```
def test_task_default_parameters(tmp_directory, tmp_imports):
    Path('script.py').write_text("""
# %% tags=["parameters"]
upstream = None
product = None

# %%
1 + 1
""")
    spec = DAGSpec({
        'task_defaults': {
            'NotebookRunner': {
                'check_if_kernel_installed': False
            }
        },
        'tasks': [{
            'source': 'script.py',
            'product': 'report.html'
        }]
    })

    dag = spec.to_dag()
    assert dag['script'].check_if_kernel_installed is False
```

I saw that in `DAGSpec._to_dag()` each task gets processed through `dagspec.py/process_tasks`. Inside of `process_tasks()`, I saw that each task gets initialized through `task, up = task_dict.to_task(dag)`. I am having trouble understanding how to access the `task_defaults` within this `to_task()` to pass onto `_init_task`. I see the method might not have access to `task_defaults` at the moment so I tried changing the call to `to_task()` within `process_tasks` to pass the `task_defaults`

```
if dag_spec['task_defaults'] is not None:
            task, up = task_dict.to_task(dag, dag_spec['task_defaults'])
        else:
            task, up = task_dict.to_task(dag)
```
and adjusting `to_task()` as shown below.

```
    def to_task(self, dag, taskdefaults=None):
...
        else:
            return _init_task(data=data,
                              meta=self.meta,
                              project_root=self.project_root,
                              lazy_import=self.lazy_import,
                              dag=dag,
                              task_defaults=task_defaults), upstream
```
This almost got me there as I now have the `task_defaults` within the `_init_task` I worked on. However, this throws a KeyError exception for NotebookRunner as I believe it expects a dict and this is a class. 

```
        if task_defaults:
>           init_params = {**task_defaults[class_], **task_dict}
E           KeyError: <class 'ploomber.tasks.notebook.NotebookRunner'>
```
Any help on how I can pass these defaults to `_init_task` in a way I can handle them as shown above would be greatly appreciated. 😄

edublancas:
I think the problem is that the user is submitting a dictionary with strings as keys (in your case `NotebookRunner`, but you're passing the class object). You can get the name of a class with `class_object.__name__`, which should return the `NotebookRunner` string.

Also, check my comment on the [original issue](https://github.com/ploomber/ploomber/issues/648#issuecomment-1073143121), I added a few items I realized were missing

jramirez857:
Thanks for pointing me back to the original issue, @edublancas . There was some useful information there that I had forgotten about. I have just pushed some code and have passed the test case I had previously.  Could you check if we are on the same page about the usage of this feature?

Looking back at the previous comments you mentioned I should cover a couple scenarios.

 

- [ ] test that one can set the task_defaults and the task is initialized with them
- [ ] test that task_defaults also work with grid
- [ ] test that task_defaults validates the keys in the dictionary (they must be task classes)

I believe I have covered the first one here with the scenario if the use case fits. I can work on implementing the `task_defaults` working with grid as well as the validation. Where can I get a good idea of how this should work with the grid functionality? Are there any useful tests you can point me to? Also, what should I be validating in terms of the classes? i.e. Should I be validating that the keys passed in for each class have valid attributes?

edublancas:
> Looking back at the previous comments you mentioned I should cover a couple scenarios.

the test cases look right

> Where can I get a good idea of how this should work with the grid functionality?

there's an if statement where we initialize the tasks that goes one way if the user has the `grid` keyword and the other way if it doesn't (you implemented the latter). so now you need to implement it in the other side of the `if` statement. this one should be simple now that you have the other implementation since `grid` passes the arguments to a `TaskGrid` (I believe that's the name of the class),  you can pass the defaults there

> Also, what should I be validating in terms of the classes? 

the validation should be the names of the task classes, I believe taskspec.py already has the list of task classes since we need to validate them in the `class:` key. For example: `PythonCallable` is valid task class name, but `some_random_string` is not. if the user passes a string that's not a valid task class name, it should say so, example "some_random_string is not a valid task class name, pass one of: (list of valid values)"

jramirez857:
I apologize for the delay. I will get back to this later this week.

Issue #472, opened 2022-01-08
confusing error if wrong extension


edublancas:
ploomber infers the type of task with a combination of the source and product extension, but if there's a typo a confusing error shows. example, if the user wants to dump data:

```yaml
  
  - source: another.sql 
    product: another.parquets # typo here
```
error:

```pytb
ploomber.exceptions.SourceInitializationError: Error initializing SQLScriptSource('SELECT *\nFROM nu...ams"], "param") ]]') Loaded from: /Users/Edu/Desktop/sql/another.sql: The {{product}} placeholder is required. Example: 'CREATE TABLE {{product}} AS (SELECT * FROM ...)'
```

Fix: see if the value in `product` has something similar to a `.parquet` or `.csv` extension and show an error

idomic:
@lgfunderburk another straightforward one :)

idomic:
@jramirez857 Another good first one

jramirez857:
Sure, I can take this up!

neelasha23:
@edublancas Is this issue sill open? Can I take it up?

idomic:
I think @jramirez857 already started working on this

jramirez857:
@neelasha23 feel free to take this! I am currently working on a different issue.

edublancas:
@neelasha23 as a pointer, you can use [this](https://docs.python.org/3/library/difflib.html#difflib.get_close_matches) to compute string distance to see if the user had a typo (e.g. parque instead of parquet)

neelasha23:
Sure! Starting on this..

neelasha23:
@edublancas 
I run the SQL template pipeline : https://github.com/ploomber/projects/tree/master/templates/spec-api-sql
After changing the extension like so:
```
- class: SQLDump
    source: join_dump.sql
    product: output/joined_data.cv
    name: join_dump
    chunksize: null
```
the pipeline still runs successfully and creates `output/joined_data.cv` which has proper data. 

The same thing happened when i changed .parquet to .parquets (https://github.com/ploomber/projects/tree/master/guides/first-pipeline)  ; in fact the pipeline still runs successfully with proper output data even if I remove the .parquet extension completely.

Am i missing something here?

edublancas:
the example is passing `class: SQLDump`, which tells Ploomber the type of task that is, please remove that line:

```yaml
-  source: join_dump.sql
    product: output/joined_data.cv
    name: join_dump
    chunksize: null
```

[...]

PR #674, opened 2022-03-22
Product extension validation - #472

neelasha23:
The code changes aim to validate product extension against csv, parquet in Sql scripts and raise the relevant error message.

### DIFF
--- a/src/ploomber/spec/taskspec.py
+++ b/src/ploomber/spec/taskspec.py
@@ -47,6 +47,13 @@ def _looks_like_path(s):
         return '/' in s
 
 
+def _extension_typo(extension, valid_extensions):
+    try:
+        return get_close_matches(extension, valid_extensions)
+    except Exception:
+        return None
+
+
 def task_class_from_source_str(source_str, lazy_import, reload, product):
     """
     The source field in a DAG spec is a string. The actual value needed to
@@ -70,10 +77,17 @@ def task_class_from_source_str(source_str, lazy_import, reload, product):
         partial(dotted_path.load_dotted_path, raise_=True, reload=reload))
 
     if extension and extension in suffix2taskclass:
-        if extension == '.sql' and _safe_suffix(product) in {
-                '.csv', '.parquet'
-        }:
-            return tasks.SQLDump
+        if extension == '.sql':
+            if _safe_suffix(product) in {'.csv', '.parquet'}:
+                return tasks.SQLDump
+            else:
+                possibilities = _extension_typo(_safe_suffix(product),
+                                                ['.csv', '.parquet'])
+                if possibilities:
+                    raise DAGSpecInitializationError(
+                        f'{_safe_suffix(product)!r} is not a valid extension. '
+                        f'Did you mean: '
+                        f'{", ".join(sorted(possibilities))}')
 
         return suffix2taskclass[extension]
     elif _looks_like_path(source_str):
@@ -417,22 +431,6 @@ def _init_task(data, meta, project_root, lazy_import, dag):
     return task
 
 
-def _validate_product_extension(product_raw):
-
-    # check if product extension is valid for SQL tasks
-    if isinstance(product_raw, str):
-        valid_extensions = ['parquet', 'csv']
-        product_extension = product_raw.split('.')[-1]
-        possibilities = get_close_matches(product_extension,
-                                          valid_extensions)
-        if product_extension not in valid_extensions and len(
-                possibilities) > 0:
-            raise DAGSpecInitializationError(
-                f'Invalid product extension in: '
-                f'{product_raw}. Did you mean: '
-                f'{", ".join(sorted(possibilities))}')
-
-
 # FIXME: how do we make a default product client? use the task's client?
 def _init_product(task_dict, meta, task_class, root_path, lazy_import):
     """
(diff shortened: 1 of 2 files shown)


src/ploomber/spec/taskspec.py
```diff
@@ -415,6 +417,22 @@ def _init_task(data, meta, project_root, lazy_import, dag):
...
+        valid_extensions = ['parquet', 'csv']
+        product_extension = product_raw.split('.')[-1]
+        possibilities = get_close_matches(product_extension,
+                                          valid_extensions)
+        if product_extension not in valid_extensions and len(
+                possibilities) > 0:
+            raise DAGSpecInitializationError(
+                f'Invalid product extension in: '
+                f'{product_raw}. Did you mean: '
+                f'{", ".join(sorted(possibilities))}')
```
edublancas, line 433: looks good!

src/ploomber/spec/taskspec.py
```diff
@@ -429,6 +447,8 @@ def _init_product(task_dict, meta, task_class, root_path, lazy_import):
     """
     product_raw = task_dict.pop('product')
 
+    _validate_product_extension(product_raw)
```
edublancas, line 450: please move this call to here: https://github.com/ploomber/ploomber/blob/6e7ae7ca155749f4efaf0093b7f0d0df462de924/src/ploomber/spec/taskspec.py#L72

this is where we determine if the user wants to dump data, hence, here we check the .csv/.parquet thing - because there are cases when the user has a SQL script but wants to create a new database table instead of dumping the data locally

so we need something like:

```python
        if extension == '.sql' and _safe_suffix(product) in {
                '.csv', '.parquet'
        }:
            return tasks.SQLDump
        else _extension_typo(_safe_suffix(product)):
           raise DAGSpecInitializationError('csb isnt a valid extension, did you mean csv?')
```

src/ploomber/spec/taskspec.py
```diff
@@ -46,6 +47,13 @@ def _looks_like_path(s):
         return '/' in s
 
 
+def _extension_typo(extension, valid_extensions):
+    try:
+        return get_close_matches(extension, valid_extensions)
```
edublancas, line 52: under which conditions this would throw an exception and what exception would that be? it's better to be specific (switch the `Exception` for the concrete exception it may raise)

src/ploomber/spec/taskspec.py
```diff
@@ -69,10 +77,17 @@ def task_class_from_source_str(source_str, lazy_import, reload, product):
...
+            if _safe_suffix(product) in {'.csv', '.parquet'}:
+                return tasks.SQLDump
+            else:
+                possibilities = _extension_typo(_safe_suffix(product),
+                                                ['.csv', '.parquet'])
+                if possibilities:
+                    raise DAGSpecInitializationError(
+                        f'{_safe_suffix(product)!r} is not a valid extension. '
+                        f'Did you mean: '
+                        f'{", ".join(sorted(possibilities))}')
```
edublancas, line 90: looks good :)

tests/spec/test_dagspec.py
```diff
@@ -1884,6 +1885,41 @@ def test_lazy_load_product_level_client(tmp_directory, tmp_imports,
...
+         'product': 'another.parquets',
+         'client': 'my_testing_module.get_db_client',
+         'product_client': 'my_testing_module.get_db_client'
+     }], "'.parquets' is not a valid extension. Did you mean: .parquet")])
+def test_product_extension_typo(tasks, expected, tmp_directory, client_spec):
+    Path('script.sql').write_text("""
+    SELECT * FROM my_table
+    """)
+
+    with sqlite3.connect('my.db') as conn:
```
edublancas, line 1910: i think you can simplify this test. since the `DAGSpec` call will fail, ploomber will never actually check if the data exists or not. I think if you remove this, the test will still be ok

tests/spec/test_dagspec.py
```diff
@@ -1884,6 +1885,41 @@ def test_lazy_load_product_level_client(tmp_directory, tmp_imports,
...
+        pd.DataFrame({'x': range(5)}).to_sql('my_table', conn)
+
+    data = {'tasks': tasks}
+
+    with pytest.raises(DAGSpecInitializationError) as excinfo:
+        spec = DAGSpec(data)
+        dag = spec.to_dag()
```
edublancas, line 1917: I think `spec = DAGSpec(data)` is enough, that's where the exception is raised, I think you can remove the `to_dag()` and `build` and the test will still work

edublancas (changes requested):
great work! i just added some minor comments. thanks for your hard work!

src/ploomber/spec/taskspec.py
neelasha23: I realised it was throwing exception only when _safe_suffix returns None (no extension found) on product, e.g., product = ['raw', 'table']. So I have added a condition instead since it would be a faster operation.

tests/spec/test_dagspec.py
neelasha23: Removed this

tests/spec/test_dagspec.py
neelasha23: So I checked that only DAGSpec(data) doesnt raise the original `ploomber.exceptions.SourceInitializationError`. The minimal set of statements that raises this error is 
```
spec = DAGSpec(data)
spec.to_dag()
```

However this testcase still passes (with to_dag() and build removed):
```
with pytest.raises(DAGSpecInitializationError) as excinfo:
     spec = DAGSpec(data)
```

If you think otherwise I'll remove the to_dag() as well.

Issue #669, opened 2022-03-21
validate import_tasks_from


edublancas:
when loading the `import_tasks_from: file.yaml`, we are not validating its contents, if the file is empty, `yaml.safe_load` will return `None`, which will throw a cryptic error message:

https://github.com/ploomber/ploomber/blob/beb625cc977bcd34481608a91daddc5493e0983c/src/ploomber/spec/dagspec.py#L326

Fix:
* If `yaml.safe_load` returns None, replace it with an empty list
* If returns something other than a list, raise an error saying we were expecting a list

LawrenceCheng1570:
Hi! I am new to open source and hoping to contribute. I was wondering if I could try to fix and claim this issue? Thanks!

idomic:
Sure thing, make sure to read the contributing guide!
Feel free to submit a PR

edublancas:
thanks for your interest @LawrenceCheng1570 , here's the contribution guide https://github.com/ploomber/ploomber/blob/master/CONTRIBUTING.md

PR #686, opened 2022-03-28
Added validation checks for import_tasks_from

LawrenceCheng1570:
Issue #669
Changes:
When loading `import_tasks_from: file.yaml`
1. If `yaml.safe_load` is None, replace it with an empty list
2. If `yaml.safe_load` returns something other than a list, raise an error saying we were expecting a list

### DIFF
--- a/src/ploomber/spec/dagspec.py
+++ b/src/ploomber/spec/dagspec.py
@@ -329,6 +329,15 @@ class DAGSpec(MutableMapping):
                 imported = yaml.safe_load(
                     Path(self.data['meta']['import_tasks_from']).read_text())
 
+                if imported is None:
+                    imported = []
+
+                if not isinstance(imported, list):
+                    raise TypeError(
+                        'Expected list when loading YAML file from '
+                        'import_tasks_from: file.yaml, '
+                        f'but got {type(imported)}')
+
                 if self.env is not None:
                     (imported,
                      tags_other) = expand_raw_dictionaries_and_extract_tags(
@@ -627,6 +636,7 @@ class DAGSpecPartial(DAGSpec):
     A DAGSpec subclass that initializes from a list of tasks (used in the
     onlinedag.py) module
     """
+
     def __init__(self, path_to_partial, env=None):
         with open(path_to_partial) as f:
             tasks = yaml.safe_load(f)
--- a/src/ploomber/spec/dagspec.py
+++ b/src/ploomber/spec/dagspec.py
@@ -636,7 +636,6 @@ class DAGSpecPartial(DAGSpec):
     A DAGSpec subclass that initializes from a list of tasks (used in the
     onlinedag.py) module
     """
-
     def __init__(self, path_to_partial, env=None):
         with open(path_to_partial) as f:
             tasks = yaml.safe_load(f)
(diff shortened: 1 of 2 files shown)


edublancas:
thanks for your contribution! can you add some tests? [ please add them here](https://github.com/ploomber/ploomber/blob/773b96064856bd4d5896a4cb8f3a29e40c4f81ee/tests/spec/test_dagspec.py#L1113)

In the first case, we wanna check that an empty file does not break initializing the DAGSpec object, in the second case, we want to test that an [appropriate error is shown](https://docs.pytest.org/en/6.2.x/assert.html)

let me know if you encounter issues or have questions!

edublancas:
Hi @LawrenceCheng1570 , thanks for the contribution!

I reproduced the error. The problem is that a downstream function doesn't work with empty lists, so I think the original logic needs to change a bit

instead of this:

```python
if imported is None:
    imported = []
```

it should be:

```python
if not imported:
    path = str(self.data['meta']['import_tasks_from'])
    raise ValueError(f'expected import_tasks_from file ({path!r}) to return a list of tasks, got: {imported}')
```

Note that this conditional will take care of `None,` `[]`, and empty dictionaries.

Please also update the test so it checks that the appropriate error message is shown: [see here](https://docs.pytest.org/en/6.2.x/assert.html#assertions-about-expected-exceptions)

the condition can be something like: `assert "expected import_tasks_from" in str(excinfo.value)`

LawrenceCheng1570:
Hi @edublancas, thanks for the suggestions! It fixed my error and I just pushed my changes. Please let me know if I need to make any other edits.

src/ploomber/spec/dagspec.py
```diff
@@ -329,6 +329,18 @@ def _init(self, data, env, lazy_import, reload, parent_path,
...
+                    raise ValueError('expected import_tasks_from file '
+                                     f'({path!r}) to return a list of tasks, '
+                                     f'got: {imported}')
+
+                if not isinstance(imported, list):
+                    raise TypeError(
+                        'Expected list when loading YAML file from '
+                        'import_tasks_from: file.yaml, '
+                        f'but got {type(imported)}')
+
```
edublancas, line 343: looks good!

tests/spec/test_dagspec.py
```diff
@@ -1129,6 +1129,33 @@ def test_import_tasks_from(tmp_nbs):
...
 
 
+def test_import_tasks_from_empty_yaml_file(tmp_nbs):
+    with pytest.raises(ValueError) as excinfo:
+        Path('some_tasks.yaml').write_text('')
+
+        spec_d = yaml.safe_load(Path('pipeline.yaml').read_text())
+        spec_d['meta']['import_tasks_from'] = 'some_tasks.yaml'
+
+        spec = DAGSpec(spec_d)
```
edublancas, line 1139: as a best practice when writing tests that evaluate error messages. it's best to only include the statement that throws the error, in this case:

```python
with pytest.raises(ValueError) as excinfo:
    DAGSpec(spec_d)
```

Please move everything the `spec = ...` statement outside the `with pytest.raises` line, and you can remove the `spec.to_dag().render()`

edublancas (changes requested):
almost there! just minor observations

tests/spec/test_dagspec.py
LawrenceCheng1570: Thanks @edublancas! Just pushed my changes

edublancas:
wooo great job! thanks a lot for contributing!

LawrenceCheng1570:
Thanks so much! Really appreciate all the help and support for my first time contributing :)

PR #695, opened 2022-03-31
To fix: error when finding pipeline.yaml in root directory (Issue 497)

neelasha23:


src/ploomber/util/default.py
```diff
@@ -456,6 +460,12 @@ def find_root_recursively(starting_dir=None,
         starting_dir=starting_dir,
     )
 
+    if root_by_pipeline == _filesystem_root():
+        raise DAGSpecInvalidError(
+            'pipeline.yaml cannot be in the filesystem root. '
+            'Please add it inside a directory like '
+            'project-name/pipeline.yaml. ')
```
edublancas, line 467: please use the filename to determine the error message, since it might change. 

```python
f'project-name/{filename}'
```

tests/util/test_default.py
```diff
@@ -305,6 +306,22 @@ def test_finds_setup_py(tmp_directory):
...
 
+@pytest.mark.parametrize('return_data', [[(None, None),
+                                          (default._filesystem_root(), None)],
+                                         [(default._filesystem_root(), 3),
+                                          (default._filesystem_root(), 1)]])
+def test_pipeline_yaml_in_root_directory(return_data, monkeypatch):
+    mock = Mock(side_effect=return_data)
+    monkeypatch.setattr(default, 'find_parent_of_file_recursively', mock)
+    with pytest.raises(DAGSpecInvalidError) as excinfo:
+        default.find_root_recursively()
```
edublancas, line 317: add add a test by calling `find_root_recursively` with a different `filename` like `pipeline.another.yaml`

edublancas (changes requested):
almost there! just minor changes, thanks a lot for working on this!

src/ploomber/util/default.py
neelasha23: Thanks for the review @edublancas  ! Made the changes

PR #704, opened 2022-04-07
Initial welcome email piece + tests

idomic:


edublancas:
looks good. I'd say we should check this in two places: `ploomber install` and `ploomber build` as there might be users that use one or the other. but ensure that if we showed it during install, we don't show it again during a build. and only show it one time

setup.py
```diff
@@ -85,7 +85,7 @@ def read(name):
     # needed to run some test pipelines
     'matplotlib',
     'seaborn',
-    'psycopg2',
+    'psycopg2-binary',
```
edublancas, line 88: why the change? psycopg2 works fine, the binary is only recommended when it isn't possible to install from source

src/ploomber/cli/cloud.py
```diff
@@ -25,6 +25,7 @@
 
 CLOUD_APP_URL = 'ggeheljnx2.execute-api.us-east-1.amazonaws.com'
```
edublancas, line 26: I think this should be the api.ploomber.io, right?

src/ploomber/cli/cloud.py
```diff
@@ -243,3 +244,44 @@ def wrapper(*args, **kwargs):
...
+    # Validate that's the first email registration
+    user_conf_path = Path(check_dir_exist(CONF_DIR), DEFAULT_USER_CONF)
+    conf = read_conf_file(user_conf_path)
+    email = conf.get('user_email', None)
+    if not email:
+        email = _get_input(
+            "Our users enjoy updates, support and funtivities "
+            "through email, please add your email would you like "
+            "to register (type email): ")
+        _email_validation(email)
```
edublancas, line 263: this will prompt the user every time, right? I think it should be a one-off thing

edublancas (changes requested):

edublancas:
one more thing: I think we should either add a timeout or only ask for the email when using interactive commands like `ploomber examples`, or `ploomber interact`, because `ploomber build` runs in production at several places and this will get stuck

setup.py
idomic: Does the psycopg2 works when developing? It caused a lot of issues that's why I changed it, but I can revert

src/ploomber/cli/cloud.py
idomic: Yeah it was tested with the direct api

src/ploomber/cli/cloud.py
idomic: No it'll be a one time off, check the _email_input func

setup.py
edublancas: I also had issues with the `binary` I think it's because there aren't binaries for all platforms. I think let's leave the original one and add a suggestion to the CONTRIBUTING.md that they may also install the binary if this one has issues

src/ploomber/cli/cloud.py
edublancas: Can you add a test? I tried it and it prompted me twice in a row

idomic: Added

setup.py
idomic: Yup makes sense

idomic:
Moved to a new cleaner PR

PR #720, opened 2022-04-18
Refactors config files

edublancas:
I refactored the logic to store and retrieve the data from the config files. This is a pre-requisite to start working on the improved onboarding workflow where we have a "tutorial" to point to the user what do to next since we need to add a few extra flags to the config files to track the state.

The tests remain unchanged (I only modified a few and deleted some that are outdated)

### DIFF
--- a/src/ploomber/config.py
+++ b/src/ploomber/config.py
@@ -29,7 +29,7 @@ class Config(abc.ABC):
             try:
                 content = yaml.safe_load(text)
                 loaded = True
-            except Exception as e:
+            except Exception:
                 warnings.warn(f'Error loading {str(path)!r}, '
                               'reverting to default values')
                 loaded = False
--- a/src/ploomber/config.py
+++ b/src/ploomber/config.py
@@ -1,3 +1,4 @@
+import warnings
 import abc
 from collections.abc import Mapping
 
@@ -6,6 +7,8 @@ import yaml
 
 class Config(abc.ABC):
     def __init__(self):
+        self._init_values()
+
         # resolve home directory
         path = self.path()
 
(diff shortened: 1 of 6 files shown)


edublancas:
can you review this @idomic ?

src/ploomber/config.py
```diff
@@ -0,0 +1,110 @@
+import warnings
+import abc
+from collections.abc import Mapping
+
+import yaml
+
+
+class Config(abc.ABC):
+    """And abstract class to create configuration files (stored as YAML)
```
idomic, line 9: Typo

src/ploomber/telemetry/telemetry.py
```diff
@@ -57,6 +55,35 @@
...
+    version_check_enabled: bool = True
+    cloud_key: str = None
+    stats_enabled: bool = True
+
+    @classmethod
+    def path(cls):
+        return Path(check_dir_exist(CONF_DIR), DEFAULT_USER_CONF)
+
+
+class Internal(Config):
```
idomic, line 70: Maybe call it ploomber settings?

tests/telemetry/test_telemetry.py
```diff
@@ -102,24 +144,18 @@ def test_env_var_takes_precedence(monkeypatch,
...
 
-    stats = Path('stats')
-    stats.mkdir()
-
-    # This isn't a first time usage since the config file doesn't exist yet.
-    assert not telemetry.check_first_time_usage()
-    (stats / 'config.yaml').write_text("stats_enabled: True")
-
     assert telemetry.check_first_time_usage()
+    assert not telemetry.check_first_time_usage()
```
idomic, line 159: How does this pass, now that we don't set the conf?

idomic (changes requested):
Looks good besides comments

src/ploomber/config.py
edublancas: fixed

src/ploomber/telemetry/telemetry.py
edublancas: I want to be more explicit that these are for internal use and others are for users.

tests/telemetry/test_telemetry.py
edublancas: I added a one-off flag to the config to replicate this functionality

idomic: Cool just make sure it works via the dashboard

idomic (approved):
Just make sure to test/squash commits.
Besides, lgtm

PR #729, opened 2022-04-26
New email, merged mainline

idomic:


edublancas:
I ran "ploomber examples" a few times, saw the email prompt and typed enter (did not enter email). Every time, I was asked for an email. But this shouldn't happen (we should only ask once). We can use the new config to set a one-time flag to do this: 

https://github.com/ploomber/ploomber/blob/5a300ab6bce131ef08a398707bc85e0f3ecc2c5f/src/ploomber/telemetry/telemetry.py#L266

I noticed that after submitting an email, the "ploomber examples" have a slight delay. Is it hitting the lambda again?

Finally, I think there should be a newline character to separate the regular "ploomber examples" output from the email signup, looks too cluttered

```
To run these examples in free, hosted environment, see instructions at: https://github.com/ploomber/projects
To download: ploomber examples -n name -o path
Example: ploomber examples -n templates/ml-basic -o ml
Our users enjoy updates, support and unique content through email, please add your email, if you'd like to register (type email): 
```

idomic:
> I ran "ploomber examples" a few times, saw the email prompt and typed enter (did not enter email). Every time, I was asked for an email. But this shouldn't happen (we should only ask once). We can use the new config to set a one-time flag to do this:
> 
> https://github.com/ploomber/ploomber/blob/5a300ab6bce131ef08a398707bc85e0f3ecc2c5f/src/ploomber/telemetry/telemetry.py#L266
> 
> I noticed that after submitting an email, the "ploomber examples" have a slight delay. Is it hitting the lambda again?
> 
> Finally, I think there should be a newline character to separate the regular "ploomber examples" output from the email signup, looks too cluttered
> 
> ```
> To run these examples in free, hosted environment, see instructions at: https://github.com/ploomber/projects
> To download: ploomber examples -n name -o path
> Example: ploomber examples -n templates/ml-basic -o ml
> Our users enjoy updates, support and unique content through email, please add your email, if you'd like to register (type email): 
> ```

Fixed, there was a missing test for empty emails. There's already a flag used called settings.user_email

Issue #120, opened 2020-05-10
Run doctests with pytest


edublancas:
https://docs.pytest.org/en/latest/doctest.html

Having some issues to make this work, might be related to: https://github.com/pytest-dev/pytest/issues/2854

shivesh01:
could you please assign this to me?

idomic:
@shivesh01 any updates here?

PR #731, opened 2022-04-30
Added test to dag.py

snooproc:
create a simple test


idomic:
@snooproc is this still work in progress?

edublancas:
For context: this is related to https://github.com/ploomber/ploomber/issues/120

A few months ago, I tried running inline examples but pytest failed to run them. During the PyCon sprint @snooproc was able to run this example test so we confirm it was a problem on my end. There's no need to merge this, but we should take another stab #120

Issue #639, opened 2022-03-06
add ploomber cloud {cmd} cli suggestions


edublancas:
When a user types an incorrect command:

```
ploomber {cmd}
```

We print a "did you mean {another}?" if there's a close match, but this functionality does not work with nested commands (e.g. `ploomber cloud`)

aadityasinha-dotcom:
Hey!! Can I work on this issue?

PR #747, opened 2022-05-10
add ploomber cloud nested suggestion for cli

bppont:
Previously, when a user input an incorrect nested command for `ploomber cloud`, ploomber cli didn't suggest a close match for the nested command. This PR enables nested command suggestion for the ploomber cloud command options (ie get-key, get-pipelines, etc)

2 files changed - 1 includes the updated cli functionality and added 2 new tests.

Issue for reference: https://github.com/ploomber/ploomber/issues/639


edublancas:
great solution! thanks for your contribution!

PR #750, opened 2022-05-10
add print feedback when running ploomber task

jrsecret:


idomic:
@jrsecret can you add a few tests?

src/ploomber/cli/task.py
```diff
@@ -63,6 +63,8 @@ def _task_cli(accept_task_id=False):
                 api.tasks_update(getattr(args, 'task_id'), 'failed')
             raise
         else:
+            print("{} executed successfully!".format(task.name))
```
edublancas, line 66: please change this to use `click.echo` it works better across operating systems, other than that, it looks good (just add some tests)

edublancas (changes requested):

src/ploomber/cli/task.py
jrsecret: Thanks for the feedback! I just changed it.

PR #755, opened 2022-05-12
change extension in pipeline.yaml when changing scripts/notebooks format

idomic:
As part of the inner task extension format, for each change we're editing the pipeline.yaml file


idomic:
A bit more efficient solution can be reiterating on the changed tasks after change and then reading/writing to the file once, but then we have to maintain the state of the old extensions for changed files.

src/ploomber/sources/notebooksource.py
```diff
@@ -606,6 +606,9 @@ def format(self, fmt):
         if extension_changed:
             path = self._path.with_suffix(ext_format)
             Path(self._path).unlink()
+            modified_entry = (Path(entry_point).read_text().replace(
```
edublancas, line 609: there's an edge case here. we also support other types of entry points, for example, it might be a path_to_module.path_to_function string so `read_text()` will fail in that case.

I suggest checking `Path(entry_point).is_file()`, if False, we should print a message with `click.secho` (in yellow font) saying that the entry point is not a file, and skip the extension replacement - the user will have to do it manually.

src/ploomber/sources/notebooksource.py
```diff
@@ -606,6 +606,9 @@ def format(self, fmt):
...
+                f'{self.name}{ext_file}', f'{self.name}{ext_format}'))
```
edublancas, line 610: in the rare event that `{self.name}{ext_file}` does not appear in the pipeline.yaml content, I think we should show a warning message as well (click.secho). this might happen if the user has an `env.yaml` file

```yaml
path: script.py
```

and in the pipeline.yaml, it has the `{{script}}` placeholder.

edublancas: to test that the warning message appears you can use capsys https://docs.pytest.org/en/6.2.x/capture.html

and pass the `--entry-point test_pkg.entry.plain_function` argument

edublancas (changes requested):

src/ploomber/sources/notebooksource.py
idomic: Functions doesn't have format so it doesn't even gets to this code, it throws an exception
Out[1]: PythonCallableSource(function) (defined at: '/private/var/folders/qx/px2tf0rn5bq6c79yd3gk37fwzpvmb9/T/tmpgnur3y2z/content/my_tasks/raw/functions.py:4')
'PythonCallableSource' object has no attribute 'format' We can either implement it with the secho or print it as part of the except block

edublancas: I didn't mean a function which is a task in a pipeline. I meant an entry point. We support users defining a function that returns a DAG object. Example:

```
ploomber build --entry-point some_module.some_function
```

In such case, the code will break because entry_point isn't a file.

Example here: https://docs.ploomber.io/en/latest/user-guide/parametrized.html#Python-API-(factory-functions)

tests/cli/test_nb.py
```diff
@@ -82,6 +82,47 @@ def test_format_skips_non_notebooks(monkeypatch, backup_simple,
...
+    monkeypatch.setattr(sys, 'argv', ['ploomber', 'nb', '--format', 'py'])
+    pipeline = Path('pipeline.yaml').read_text()
+    cli.cmd_router()
+
+    assert pipeline == Path('pipeline.yaml').read_text()
+
+
+def test_format_no_entry_point(monkeypatch, tmp_nbs_factory, capsys):
+    monkeypatch.setattr(sys, 'argv', [
+        'ploomber', 'nb', '--entry-point', 'factory.make', '--format', 'ipynb'
```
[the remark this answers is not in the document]
edublancas: I think `factory.make` isn't working because the module isn't in the path. so if you go with `test_pkg` you can delete the `nbs-factory` code and fixture.

edublancas (changes requested):

PR #769, opened 2022-05-16
Adds terminal styling to Python inline examples (fixes #766)

### LINKED ISSUE #766: enable copy paste for inline examples
We need to add some CSS and JS so the Python inline examples (the ones that start with `>>>` can be copied). This is how a non-Python example looks like:

image

If the user clicks on the window, it'll copy the code. This is how a Python example looks like:

image

We should add the window-like CSS and the copy functionality - the CSS and JS is already there, we just need to connect them with these other types of examples.

Source: https://docs.ploomber.io/en/latest/api/_modules/tasks/ploomber.tasks.PythonCallable.html#ploomber.tasks.PythonCallable

CouchFerret:
## Changes:

- Python inline examples in the source code have been moved into code-blocks
- Terminal style has been added to these code-blocks by their `python-console` class name

## Decisions:

- I decided to move the Python inline examples into code-blocks, because without it they only get the default code-block css classes after doc gen, which makes it difficult to select them and apply the terminal style. Without this I either have to apply the styling to all default code-blocks or select them by their unique structure/content (like, looking for child elements with the content of `>>> `).

### DIFF
--- a/doc/_static/js/custom.js
+++ b/doc/_static/js/custom.js
@@ -18,6 +18,9 @@ $(document).ready(function () {
     editor = document.getElementsByClassName("text-editor");
     addTerminalStyle(editor, "Text editor", (buttons = true));
 
+    python = document.getElementsByClassName("python-console");
+    addTerminalStyle(python, "Terminal (Python)", (buttons = true));
+
     // TODO: this should listen for click events in the whole terminal
     // window, not only in the content sub-window
     $("div.highlight").click(function () {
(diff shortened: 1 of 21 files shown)


edublancas:
Hi, thanks for your contribution!

Is there any other way to select them? The issue with embedding these into a `code-block` is that contributors may forget to add this and we'll end up with some chunks that have the copy feature and others that don't. 

perhaps a selector that takes into account that all these snippets are inside a `dl.py.class` ?

```css
dl.py.class > div.highlight {
}
```

I'm pretty bad at CSS so I'll leave it up to you to tell me if this is a good idea.

edublancas:
closing due inactivity, feel free to re-open if you wanna work on it

PR #775, opened 2022-05-17
Improved error message

### LINKED ISSUE #642: Improve error message when task function has wrong signature
When a user declares a function with the argument `products` (instead of `product`), the error looks like this:

```
Error rendering task "get" initialized with function "get". Got unexpected arguments: ['product']. Missing arguments: ['products']. Pass ['products'] in "params"
```

This is confusing because the user may end up passing `products` under the params section. But under this scenario, the user really meant to pass `product`.

Solution: check the extra arguments, and see if there is one that is similar to `product` to detect typos, then modify the error message so it says something like:

```
You are passing products. Did you mean product?
```

We can use difflib to get close matches

aadityasinha-dotcom:
Improved error message whenever a user passes `products` instead of `product` using `difflib`.
Resolves #642

### DIFF
--- a/tests/util/test_util.py
+++ b/tests/util/test_util.py
@@ -22,6 +22,8 @@ def get(products):
 
 def test_signature_check():
     error = ("You are passing products. Do you mean product?")
+
     with pytest.raises(TaskRenderError) as e:
         signature_check(get, {'product': 1}, 'get')
+        
     assert error == str(e.value)
--- a/tests/util/test_util.py
+++ b/tests/util/test_util.py
@@ -1,0 +1,23 @@
+import inspect
+from sklearn import datasets
+from pathlib import Path
+import pandas as pd
+
+from ploomber.util.util import signature_check
+
+def get(products):
+    """Get data
+    """
+    d = datasets.load_iris()
+
+    df = pd.DataFrame(d['data'])
+    df.columns = d['feature_names']
+    df['target'] = d['target']
+
+    Path(str(product)).parent.mkdir(exist_ok=True, parents=True)
+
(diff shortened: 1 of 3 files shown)


edublancas:
can you add a test?

aadityasinha-dotcom:
> can you add a test?

Yeah sure!!

tests/util/test_util.py
```diff
@@ -0,0 +1,23 @@
...
+    df.columns = d['feature_names']
+    df['target'] = d['target']
+
+    Path(str(product)).parent.mkdir(exist_ok=True, parents=True)
+
+    df.to_parquet(str(product))
+
+def test_signature_check():
+    x = signature_check(get, inspect.signature(get).parameters, 'get')
+    assert x == True
```
edublancas, line 23: this test is not relevant. please test the changes that you introduced. in other words, that if the function has a `products` argument, you get the new error suggesting to change it to `product`

edublancas (changes requested):

aadityasinha-dotcom:
@edublancas I have made the changes.

edublancas (changes requested):

edublancas:
correct!

aadityasinha-dotcom:
@edublancas how can I test the pipeline as `signature_check` is verifying the function not the pipeline I tried using `PythonCallableSource` but didn't get the expected error.

edublancas:
using PythonCallable should work, upload the test that you tried to see what's missing

aadityasinha-dotcom:
okay

aadityasinha-dotcom:
Was the PR incorrect?

PR #815, opened 2022-05-26
Better error message when doing cloud status

### LINKED ISSUE #808: better error message when pipeline is up-to-date
this message is a bit confusing:

https://github.com/ploomber/ploomber/blob/8b036af2fa12be4907cd716138ee23bb51d70d2f/src/ploomber/cloud/api.py#L178

I think we should check if the pipeline status is "finished", in such case tell the user that the pipeline finished. Also, if the pipeline finished because the incremental builds did not trigger any new run, we should mention it, and suggest the user run `ploomber cloud build --force`

Wxl19980214:
Fixed #808 with a more explicit error message.
Added test case.

### DIFF
--- a/src/ploomber/cloud/api.py
+++ b/src/ploomber/cloud/api.py
@@ -91,6 +91,7 @@ def download_from_presigned(presigned):
 
 
 def auth_header(func):
+
     @wraps(func)
     def wrapper(*args, **kwargs):
         api_key = get_key()
@@ -172,10 +173,24 @@ def run_detail_print(run_id):
 
     if run['status'] == 'created':
         click.echo('Run created...')
+    elif run['status'] == 'finished':
+        click.echo('Pipeline finished...')
+        if tasks:
+            click.echo(Table.from_dicts(tasks))
+        else:
+            click.echo('Pipeline finished due to no newly triggered tasks,'
+                       ' try running ploomber cloud build --force')
     elif tasks:
+        if run['status'] == 'aborted':
+            click.echo('Pipeline aborted...')
+        elif run['status'] == 'failed':
+            click.echo('Pipeline failed...')
+        else:
+            click.echo('Unknown status: ' + run['status'])
         click.echo(Table.from_dicts(tasks))
     else:
-        click.echo('Pipeline up-to-date, no tasks scheduled for this run.')
+        click.echo('Unknown status: ' + run['status'] +
+                   ', no tasks triggered.')
 
     return out
 
--- a/tests/cli/test_cloud_execution.py
+++ b/tests/cli/test_cloud_execution.py
@@ -42,7 +42,7 @@ def test_cloud_status(runid):
     assert 'name' in result.output
     assert 'runid' in result.output
     assert 'status' in result.output
-    assert len(result.output.splitlines()) == 16
+    assert len(result.output.splitlines()) == 17
     assert result.exit_code == 0
 
 
--- a/tests/cloud/test_api.py
+++ b/tests/cloud/test_api.py
@@ -100,3 +100,121 @@ def test_upload_project_ignores_product_prefixes(monkeypatch, tmp_nbs):
         files = zip.namelist()
 
     assert 'output/should-not-appear' not in files
+
+
+def test_run_detailed_print_finish_no_task(monkeypatch, capsys):
+
+    def mock_return(self):
+        return {'run': {'status': 'finished'}, 'tasks': []}
+
+    monkeypatch.setattr(api, 'run_detail', mock_return)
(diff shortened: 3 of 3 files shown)


tests/cli/test_cloud_execution.py
```diff
@@ -46,6 +46,32 @@ def test_cloud_status(runid):
     assert result.exit_code == 0
 
 
+def test_cloud_status_finish_with_no_task_run(monkeypatch, runid):
+
+    class MockResponse:
```
idomic, line 51: You can just use Mock() with a return value, no need here to create a class.

tests/cli/test_cloud_execution.py
```diff
@@ -46,6 +46,32 @@ def test_cloud_status(runid):
...
+                    try build with --force'
+
+    def mock_status(*args, **kwargs):
+        return MockResponse()
+
+    monkeypatch.setattr(CliRunner, 'invoke', mock_status)
+    runner = CliRunner()
+    result = runner.invoke(cli, ['cloud', 'status', runid])
+
+    assert 'taskid' not in result.output()
```
idomic, line 67: The assertions here are off, what we're looking to test is the user output in the switchcase, make sure to use the capsys module and read the out pipe and assert on it. No need to assert stuff on the mock since it doesn't test anything.

idomic (changes requested):
Please fix the comments

src/ploomber/cloud/api.py
```diff
@@ -175,7 +175,11 @@ def run_detail_print(run_id):
     elif tasks:
         click.echo(Table.from_dicts(tasks))
     else:
-        click.echo('Pipeline up-to-date, no tasks scheduled for this run.')
+        if run['status'] == 'finished':
```
idomic, line 178: What happens here on aborted pipelines?

idomic: Also can you move those conditions a level above to being part of the switchcase?

idomic (changes requested):
A few comments on the condition statement

idomic:
@Wxl19980214 please make sure to lint and that the CI is passing.

src/ploomber/cloud/api.py
```diff
@@ -170,12 +171,25 @@ def run_detail_print(run_id):
...
+        # both failed and aborted pipeline contain tasks
+        click.echo('Pipeline aborted...')
+        click.echo(Table.from_dicts(tasks))
+    elif run['status'] == 'failed':
+        click.echo('Pipeline failed...')
         click.echo(Table.from_dicts(tasks))
     else:
-        click.echo('Pipeline up-to-date, no tasks scheduled for this run.')
+        click.echo('Pipeline finished...')
+        if tasks:
```
idomic, line 186: We can take this a level up and remove 183/180:
elif tasks:
  click.echo(Table.from_dicts(tasks))

src/ploomber/cloud/api.py
```diff
@@ -170,12 +171,25 @@ def run_detail_print(run_id):
...
+            # normal finish, print tasks
+            click.echo(Table.from_dicts(tasks))
+        else:
+            # pipeline does not run at all
+            click.echo('Pipeline finished due to no newly triggered tasks,'
+                       ' try build with --force')
```
idomic, line 192: on 192, try running: `ploomber cloud build --force`.

Also, what's separating it from a regular finish? like if there's a pipeline that finished as expected, I believe it'll fall into the else clause right?

idomic (commented):
please fix changes

src/ploomber/cloud/api.py
Wxl19980214: I thought there are two situation for finishing? 1. regular finish with tasks, we print out task information on line 188. 2. finish because of no new tasks run, there will be no tasks, so we go to line 191 print the message?

idomic: Yes you're right, the regular finish will print the tasks and a message of pipeline finished in this case.

src/ploomber/cloud/api.py
edublancas: maybe for completeness also include  the `click.echo(Table.from_dicts(tasks))` in the finished case.

I'd also suggest explicitly doing `== 'finished'` and leave the else statement for an unknown case:

```
else:
   click.echo('Unknown status: {status}')
   click.echo(Table.from_dicts(tasks))
```

idomic:
@Wxl19980214 Please make sure the CI is passing.

idomic:
Nice job! We could improve here on the process, try self code-review first make sure you're adding tests and documentation before pushing 👍

PR #817, opened 2022-05-27
Added tests for 3.10

idomic:


edublancas:
ah yeah. I see them. it's because they need pygraphviz. I think let's only skip them when running on Python 3.10 and windows (so both conditions)

idomic:
pygraphviz had issues with 3.10 so it's disabled on linux and macos as
well, so the current condition is <3.10, I've added those skips lets see if
it's passing now.

edublancas:
got it!

edublancas:
the CI is failing but I think it's because of the intermittent issue. lmk when it's ready for review

edublancas:
i wanted to trigger an extra run to see the CI, will merge shortly

Issue #805, opened 2022-05-22
add did you mean feature to `ploomber examples`


edublancas:
We should add the "did you mean?" feature when executing `ploomber examples`

```sh
 ploomber examples -n cookbook/fileclient -o fileclient
```

```txt
There is no example named "cookbook/fileclient", did you mean "cookbook/file-client"?
```

for reference: We already have this built-in in other places https://github.com/ploomber/ploomber/blob/907c5ed798354efbedc5f4cfd397179e75812ec5/src/ploomber_cli/cli.py#L17

aadityasinha-dotcom:
Can you explain this issue?

PR #825, opened 2022-05-31
added and tested did you mean feature to ploomber examples issue #805

qixuan27:
## Describe your changes
added and tested did you mean feature to ploomber examples issue #805

Closes #x 


tests/cli/test_examples.py
```diff
@@ -339,6 +339,18 @@ def test_error_if_git_clone_fails(monkeypatch, capsys):
...
 
 
+def test_did_you_mean_feature():
+    examples.main(name=None, force=True)
+
+    runner = CliRunner()
+    result = runner.invoke(cli.cli,
+                           ['examples', '--name', "report-generation"])
+
+    # print("this is a test")
```
idomic, line 349: Please clean all of the comments and debug code

src/ploomber/cli/examples.py
```diff
@@ -263,7 +265,7 @@ def download(self, name, output):
 
         if not selected.exists():
             raise BaseException(
-                f'There is no example named {name!r}.\n'
+                f'There is no example named {name!r}\n'
```
idomic, line 268: Why the dot is missing?

idomic (changes requested):
Please fix the comments.
Also add the issue number this closes and mark the check boxes after you perform each of those (test, self code review etc).

idomic:
@qixuan27 please make sure the CI is passing as well

src/ploomber/cli/examples.py
```diff
@@ -310,4 +312,26 @@ def main(name, force=False, branch=None, output=None):
     if not name:
         manager.list()
     else:
-        manager.download(name=name, output=output)
+        # manager.download(name=name, output=output)
```
idomic, line 315: Please remove the commented debug

idomic:
@qixuan27 please fix the ci errors

qixuan27:
Based on the error log, all AssertionErrors are in two files:
1) All AssertionErrors in test_cloud_execution.py is due to the missing API keys in result.output. I was stuck with this problem because I did not modify the part of code this file is testing for.
2) All AssertionErrors in test_examples.py is due to python version issues. These tests are passed well locally on my PC with python 3.9. However, it failed the test for python 3.6 on the various OS repeatedly. I tried my best to solve these version issues but still have not found a way yet.

idomic:
> test_cloud_execution.py

You can ignore these errors for now (I thought I saw a different error).

> 3.6 is failing

You can have a conda env configured to python 3.6 and debug this locally

idomic:
@qixuan27 
A few pointers:
You had this line in the beginning of your test `examples.main(name=None, force=True)`
This calls the examples without any name and forces it to get to the local. This isn't necessary.

Also, after removing this line, it goes into the else statement, then in the manager.download it gets into this if: `src/ploomber/cli/examples.py:265` and throws an exception (ploomber.exceptions.BaseException), so the assertion and the rest of the test isn't really running. probably something that's missing in your code.

To debug this I've used break points and pdb, then ran the test, that's something you can also do for next time.
Try fixing the code so what we expect to happen (did you mean this example?) works with the test.

idomic:
@qixuan27 The CI will fail on lints

idomic:
@qixuan27 looks like the CI is failing, I think the easiest way is to start a new PR from mainline and migrate your code there since a lot changed in the last few weeks.

qixuan27:
> @qixuan27 looks like the CI is failing, I think the easiest way is to start a new PR from mainline and migrate your code there since a lot changed in the last few weeks.

Yes, sounds great.
I have solved the version control problems yesterday and there is only one failed test(test_does_not_download_again_if_no_explicit_branch_requested) left to be solved. So I think I can continue this PR. I will pay more attention to version control for my future submissions.

94rain:
Looks like [the CI](https://github.com/ploomber/ploomber/runs/6985541153?check_suite_focus=true) fails on an intermittent test `test_task_with_client_and_metaproduct_isnt_outdated_rtrns_waiting_download` mentioned in #847. Can we try to re-run it?

edublancas:
i think they're all passing now, except for the integration tests. (i don't know why they appear as skipped)

idomic:
@94rain please open an issue on the missing guide (I don't think it's the category since the rest are working).
Sounds good, please commit and lmk once ready for review.

tests/cli/test_examples.py
```diff
@@ -166,7 +166,9 @@ def test_does_not_download_again_if_no_explicit_branch_requested(
 
     # if called again but no force nor branch arg, it shouldn't download again
     examples.main(name=None, force=False, branch=None)
-    examples.main(name='templates/ml-online', force=False, branch=None)
+    examples.main(name=None, force=False)
+    examples.main(name=None, branch=None)
+    examples.main(name=None, force=False, branch=None)
```
edublancas, line 171: do we need these extra lines?

the original version only had `examples.main(name=None, force=False, branch=None)`

src/ploomber/cli/examples.py
```diff
@@ -301,13 +316,14 @@ def download(self, name, output):
 
 @command_endpoint
 @telemetry.log_call('examples')
-def main(name, force=False, branch=None, output=None):
+def main(name, branch=None, force=False, output=None):
```
edublancas, line 319: please revert the  arguments to the original order `main(name, force=False, branch=None, output=None)`

src/ploomber/cli/examples.py
```diff
@@ -259,11 +261,24 @@ def list(self):
...
+                  newline='',
+                  encoding='utf-8-sig') as f:
+            rows = list(csv.DictReader(f))
         selected = self.path_to(name)
-
+        categories = []
+        for row in rows:
+            category = row.pop('name')
+            del row['idx']
+            categories.append(category)
```
edublancas, line 273: I think let's first rename categories to `names` (I don't what I was thinking when I named that variable) 

looks like all of this is repeated in the `list` method. I think let's move all of this to the class constructor and assign the `names` (previously named `categories`) to an instance variable

```python
self._names = names
```

then, we can re-use `self._names` in `list` and `download`

src/ploomber/cli/examples.py
```diff
@@ -20,6 +20,7 @@
 from pygments.formatters.terminal import TerminalFormatter
 from pygments.lexers.markup import MarkdownLexer
 from pygments import highlight
+from ploomber_cli.cli import _suggest_command
```
edublancas, line 23: let's declare a new `_suggest_example` on this file based on the implementation of `_suggest_command`, since the latter has things that we don't need here.

```python
def _suggest_examplename: str, options):
    if not name or name in options:
        return None

    name = name.lower()

    close_commands = get_close_matches(name, options)

    if close_commands:
        return close_commands[0]
    else:
        return None
```

src/ploomber/cli/examples.py
```diff
@@ -259,11 +261,24 @@ def list(self):
...
+            categories.append(category)
         if not selected.exists():
+            closest_match = _suggest_command(name, categories)
+            # when suggested command returns None, disable did you mean feature
+            did_you_mean_message = f'Did you mean "{closest_match}"?\n' \
+                if closest_match is not None else ''
             raise BaseException(
-                f'There is no example named {name!r}.\n'
+                f'There is no example named {name!r}. '
+                f'{did_you_mean_message}'
```
edublancas, line 281: we need a conditional here. if there's not suggestion, we should print

> there is no message named... list examples... update...

but if there is a suggestion, then the message should only be

> there is no message named... did you mean ...?

edublancas (changes requested):

src/ploomber/cli/examples.py
94rain: I handled the conditional in this way: 
```python
            did_you_mean_message = f'Did you mean "{closest_match}"?\n' \
                if closest_match is not None else ''
```
https://github.com/qixuan27/ploomber/blob/9751d3fa2874c462591f8e473fbc883a645fe85e/src/ploomber/cli/examples.py#L277-L278

Does that look OK?

src/ploomber/cli/examples.py
94rain: Sure. Have done the change

src/ploomber/cli/examples.py
94rain: Fixed

src/ploomber/cli/examples.py
94rain: How about the `name` arg in the function signature? Could this be a bit confusing?
`    def download(self, name, output):`

edublancas: let's rename categories to `example_names` then

src/ploomber/cli/examples.py
edublancas: the problem with the current logic is that If there is a suggestion the error message is too verbose. it looks like this:

```
there is not example named ...
list examples...
update local copy...
get ml example...
```

but it should be more concise, and just:

```
there is not example named ...
```

when there is *no* suggestion, the error is fine. no need to change anything:

```
list examples...
update local copy...
get ml example...
```

src/ploomber/cli/examples.py
94rain: I was trying to merge them into the class constructor but just realized that they are different. 

* In `list()`: It's a dict (key: category, value: examples in the category), which may be the reason why it is named `category`.
```python
        by_category = defaultdict(lambda: [])
        for row in rows:
            category = row.pop('category')
            del row['idx']
            by_category[category].append(row)
```
* In `download()`, it's a list of all examples:
```python
        categories = []
        for row in rows:
            category = row.pop('name')
            del row['idx']
            categories.append(category)
```
Maybe We can just keep the former `by_category` dict and move it to the constructor, and we can use it in `list()`. Also, maybe we can rename `by_category` to `examples_by_category`?

For `download()`, looks like we need another variable `self._example_names = [example["name"] for category, examples in self._examples_by_category.items() for example in examples]` in the constructor.

94rain: After finishing the above change, the `test_clones_in_home_directory` and `test_change_default_branch` tests will fail. Looks like the constructor is not called. How should we call the constructor? `monkeypatch.setattr(examples._ExamplesManager, '__init__', lambda _: None)` does not work.

https://github.com/qixuan27/ploomber/blob/master/tests/cli/test_examples.py#L104-L113

94rain:
Pending resolving the issue mentioned in https://github.com/ploomber/ploomber/pull/825#discussion_r918279191:

After finishing the above change, the `test_clones_in_home_directory` and `test_change_default_branch` tests will fail. Looks like the constructor is not called. How should we call the constructor with `monkeypatch`? `monkeypatch.setattr(examples._ExamplesManager, '__init__', lambda _: None)` does not work.

Or maybe we can just initialize an example manager? 


https://github.com/qixuan27/ploomber/blob/master/tests/cli/test_examples.py#L104-L113

idomic:
@94rain are you trying to mock the class' constructor? [look here](https://stackoverflow.com/questions/57042557/pytest-mocking-constructor-within-constructor), you'd need to mock a class and have a container within it.

edublancas:
looks like the mock is working, the problem is that the lambda arguments are incompatible:

this is the CI error

```
    @command_endpoint
    @telemetry.log_call('examples')
    def main(name, force=False, branch=None, output=None):
        """
        Entry point for examples
        """
        examples_manager = _ExamplesManager(branch=branch,
                                            force=force,
>                                           verbose=True)
E       TypeError: <lambda>() got an unexpected keyword argument 'branch'
```

for the tests to pass the lambda needs a branch argument (and the other two)

94rain:
I managed to mock the constructor. However, seems that mocking the constructor is not intended as the `self.clone()` method will not be called and will cause the `mock_run.assert_called_once_with` assertion to fail. I'm thinking about how to better address this issue.

94rain:
Alternatively, I used try-except to print the error instead of raising it. Since for these two tests, what we only care about is whether `git clone...` is called only once (which is before the file reading part that raise `FileNotFoundException`), I think this shouldn't be an issue.
```python
    try:
        examples.main(name=None, force=False, branch='custom-branch')
    except BaseException as e:
        print(e)
```

idomic:
@94rain is this ready for review?

94rain:
> @94rain is this ready for review?

Yes

edublancas (changes requested):
two small changes, when there is a suggestion. the is a trailing newline character:

```
(ploomber)  Edu@MBP Desktop/deleteme » ploomber examples -n templates/ml-basico -o ml
Loading examples...
Error: There is no example named 'templates/ml-basico'. Did you mean "templates/ml-basic"?
[newline here]
```

let's get rid of it.

when there is no suggestion, two newline breaks are missing. currently:

```
(ploomber)  Edu@MBP Desktop/deleteme » ploomber examples -n grid -o ml                                                                 
Loading examples...
Error: There is no example named 'grid'. List examples: ploomber examples
Update local copy: ploomber examples -f
Get ML example: ploomber examples -n templates/ml-basic -o ml-example
```

but it should be:

```
(ploomber)  Edu@MBP Desktop/deleteme » ploomber examples -n grid -o ml                                                                 
Loading examples...
[newline break here]
Error: There is no example named 'grid'.[newline break here]List examples: ploomber examples
Update local copy: ploomber examples -f
Get ML example: ploomber examples -n templates/ml-basic -o ml-example
```

other than that, looks good

tests/cli/test_examples.py
```diff
@@ -112,7 +112,10 @@ def test_clones_in_home_directory(monkeypatch, tmp_directory):
     # mock list, otherwise this will fail since we aren't cloning
     monkeypatch.setattr(examples._ExamplesManager, 'list', lambda _: None)
 
-    examples.main(name=None, force=False)
+    try:
```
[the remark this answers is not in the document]
94rain: This is to ignore the exception, which is similar to the following:
```python
    with pytest.raises(BaseException):
        examples.main(name=None, force=False, branch='custom-branch')
```

PR #829, opened 2022-06-01
error message when .yml found instead of .yaml

### LINKED ISSUE #659: Pipeline can't be run with env file having .yml suffix
### Description
Pipeline can't be run with `env.yml` file. It seems it won't get read at all since variables stated in the `env.yml` are unknown to the pipeline run. Renaming file to `env.yaml` solved the issue for me. 

### Replication
Files for replication of minimal example are attached [here](https://github.com/ploomber/projects/files/8303640/test.zip).

### Task
It would be nice to not be dependent on the suffix. Adjust the code, so it treats `.yaml` and `.yml` suffixes of `env` file equally.

Wxl19980214:
## Describe your changes

When users have pipeline, env file ends with .yml extension, output error message reminding them to rename to .yaml instead.

Closes #659

### DIFF
--- a/src/ploomber/cli/parsers.py
+++ b/src/ploomber/cli/parsers.py
@@ -38,6 +38,7 @@ class CustomMutuallyExclusiveGroup(argparse._MutuallyExclusiveGroup):
     A subclass of argparse._MutuallyExclusiveGroup that determines whether
     the added args should go into the static of dynamic API
     """
+
     def add_argument(self, *args, **kwargs):
         if not self._container.finished_static_api:
             if (not self._container.in_context
@@ -65,6 +66,7 @@ class CustomParser(argparse.ArgumentParser):
     manager, only after this has happened, other arguments can be added using
     parser.add_argument.
     """
+
     def __init__(self, *args, **kwargs):
         self.DEFAULT_ENTRY_POINT = default.try_to_find_entry_point()
 
@@ -125,6 +127,9 @@ class CustomParser(argparse.ArgumentParser):
                            'Use --entry-point/-e to pass a '
                            'entry point\'s location or '
                            'place it in a standard location.\n\n'
+                           'Otherwise check if your pipeline have '
+                           '.yml as extension, '
+                           'change it to .yaml instead.\n\n'
                            'Need help? https://ploomber.io/community')
 
             return self.DEFAULT_ENTRY_POINT
--- a/src/ploomber/spec/dagspec.py
+++ b/src/ploomber/spec/dagspec.py
@@ -312,7 +312,6 @@ class DAGSpec(MutableMapping):
         env = env or dict()
         path_to_defaults = default.path_to_env_from_spec(
             path_to_spec=self._path)
-
         # there is an env.yaml we can use
         if path_to_defaults:
             defaults = yaml.safe_load(Path(path_to_defaults).read_text())
@@ -322,6 +321,11 @@ class DAGSpec(MutableMapping):
 
         # there is no env.yaml
         else:
+            # wrong extension of env.yml or env.{name}.yml found
+            if default.try_to_find_env_yml(path_to_spec=self._path):
+                raise DAGSpecInitializationError(
+                    'Error: found env file ends with .yml. '
+                    'Change the extension to .yaml.')
             self.env = EnvDict(env, path_to_here=self._parent_path)
 
         self.data, tags = expand_raw_dictionary_and_extract_tags(
@@ -663,6 +667,7 @@ class DAGSpecPartial(DAGSpec):
     A DAGSpec subclass that initializes from a list of tasks (used in the
     onlinedag.py) module
     """
+
     def __init__(self, path_to_partial, env=None):
         with open(path_to_partial) as f:
             tasks = yaml.safe_load(f)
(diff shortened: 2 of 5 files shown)


src/ploomber/spec/dagspec.py
```diff
@@ -322,6 +322,12 @@ def _init(self, data, env, lazy_import, reload, parent_path,
 
         # there is no env.yaml
         else:
+            # wrong extension of env.yml or env.{name}.yml found
```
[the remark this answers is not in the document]
Wxl19980214: Working on it

src/ploomber/cli/parsers.py
```diff
@@ -125,6 +127,9 @@ def parse_entry_point_value(self):
                            'Use --entry-point/-e to pass a '
                            'entry point\'s location or '
                            'place it in a standard location.\n\n'
+                           'Otherwise check if you pipeline have '
```
idomic, line 130: your?

src/ploomber/spec/dagspec.py
```diff
@@ -322,6 +321,12 @@ def _init(self, data, env, lazy_import, reload, parent_path,
...
+            if default.try_to_find_env_yml(path_to_spec=self._path):
+                raise DAGSpecInitializationError(
+                    'Error: found env file ends with .yml, '
+                    'which is a format that we do not '
+                    'support currently. Change it to .yaml instead.')
```
idomic, line 329: We need to give here a link to the docs (for more context) probably here: https://docs.ploomber.io/en/latest/get-started/basic-concepts.html

src/ploomber/util/default.py
```diff
@@ -257,6 +257,27 @@ def entry_point_relative(name=None):
...
+    """
+
+    The purpose of this function is to check whether there are env files
+    ending with .yml format. It will return that file if it does exist,
+    otherwise return None.
+
+    This function will only be called right after path_to_env_from_spec.
+
+    """
+    # FIXME: delete this
```
idomic, line 270: remove this comment

src/ploomber/util/default.py
```diff
@@ -257,6 +257,27 @@ def entry_point_relative(name=None):
     return location_pkg or location
 
 
+def try_to_find_env_yml(path_to_spec):
```
idomic, line 260: Why don't we make it generic? pipeline.yml is also not supported in addition to env.yml

tests/spec/test_dagspec.py
```diff
@@ -320,6 +321,47 @@ def test_doesnt_load_env_in_default_location_if_loading_from_dict(tmp_nbs):
...
+      product: output.ipynb
+""")
+
+    with pytest.raises(DAGSpecInitializationError) as excinfo:
+        DAGSpec('pipeline.yaml')
+
+    assert 'Error: found env file ends with .yml' in str(excinfo.value)
+
+
+def test_error_if_load_env_yml_from_sibl_dir(tmp_directory):
```
idomic, line 343: Add same tests for pipeline.yml (try to parameterize those functions instead of duplicating).

idomic (changes requested):
Added a few comments

src/ploomber/spec/dagspec.py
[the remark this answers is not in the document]
idomic: @edublancas correct me if I'm wrong but we want to support that exact functionality for pipeline.yml right?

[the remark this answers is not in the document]
edublancas: throwing an error if there is a pipeline.yml? yeah, we want. but this PR is about env.yml, I think let's do it separately since the fix involves changes in different places

src/ploomber/spec/dagspec.py
edublancas: I don't think there's a need to show the docs, because the fix is simple, rename env.yml to env.yaml. I'd only suggest changing the error message because it's not a different format, it's just a different extension. 

```
'Error: found env file ends with .yml. Change the extension to .yaml.'
```

idomic: Yup we should isolate the issues, agree.

src/ploomber/util/default.py
idomic: let's ask @edublancas , why we don't need this check anymore?
# FIXME: delete this
    if path_to_spec is None:
        return None

Do we alert the user this path doesn't exist somewhere else? (and this is guaranteed to have some value if called).

edublancas: I wrote this code >1 a year ago 😂 so I don't fully remember. I think it's a case that will never happen anymore. but better to remove it and see if all tests pass to confirm

Wxl19980214: I think i have tried to remove it, and it caused lots of tests to fail. I will confirm again.

src/ploomber/spec/dagspec.py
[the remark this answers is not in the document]
idomic: I've opened an issue on the pipeline.yaml, let's keep it isolated to the env.yaml

src/ploomber/util/default.py
idomic: haha yeah I figured it's pretty old. Alright, we'll have to tackle it at some point, or just remove the fixme

Wxl19980214: removed two fixme. ready to merge

Issue #786, opened 2022-05-18
colab broken due to email prompt


edublancas:
we recently added a prompt for new users to submit their email; however, this doesn't work in jupyter notebooks (including colab):

```sh
# this breaks
! ploomber examples -n {example} -o {out}
```

see https://github.com/ploomber/projects/issues/37 for reference

solving it for colab is simple since it sets and env variable, we can check for it and skip the email prompt if it appears. for jupyter, we can do something similar but I don't know if jupyter sets an environment variable

*Note: I removed the README link to colab*

94rain:
I think the email prompt actually works in Colab. The other issue is not about the email prompt but about plotting.

I am able to enter my email here:

<img width="623" alt="image" src="https://user-images.githubusercontent.com/21193371/168940112-6a5e5817-f961-46ee-8998-527a701f826d.png">

After entering my email and pressing return:

<img width="646" alt="image" src="https://user-images.githubusercontent.com/21193371/168940196-f6e60b66-54d2-4cca-92c0-200062b5ef97.png">

edublancas:
oh, thanks for letting me know.

So I dug a bit more. As you mention, the email prompt works, but it's very confusing since the field isn't visible (I had to scroll down a bit until I found it) – so I think we should fix this as well.

The error you reported is coming from IPython, colab runs a very old version (5.5.0), when I tried to upload it, I saw this:

> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
jupyter-console 5.2.0 requires prompt-toolkit<2.0.0,>=1.0.0, but you have prompt-toolkit 3.0.29 which is incompatible.
**google-colab 1.0.0 requires ipython~=5.5.0, but you have ipython 7.33.0 which is incompatible.**
google-colab 1.0.0 requires tornado~=5.1.0; python_version >= "3.0", but you have tornado 6.1 which is incompatible.
Successfully installed ipython-7.33.0 prompt-toolkit-3.0.29
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv

Upgrading IPython makes the error go away, but it seems like we're stuck. If we keep the old IPython, colab complains, but if we upgrade it, colab may break.

94rain:
[...]

PR #836, opened 2022-06-06
generate requirements-colab.lock.txt to enforce specific package versions

94rain:
## Describe your changes
To resolve the colab environment inconsistency issue, we decided to generate a requirements-colab.lock.txt file to enforce specific package versions so that the example can be run smoothly.

Part of #786


idomic:
lgtm

PR #874, opened 2022-06-23
ml-basic windows bug fix

### LINKED ISSUE #870: Error when running example
Maybe this is a windows only issue? I think this was introduced when we released the support for multiple output notebooks 

From slack:

By the way, https://github.com/ploomber/projects/tree/master/templates/ml-basic report an error after ploomber build:

`Traceback (most recent call last):
  File "C:\ProgramData\Anaconda3\envs\ag10\lib\site-packages\nbformat\reader.py", line 18, in parse_json
    nb_dict = json.loads(s, **kwargs)
  File "C:\ProgramData\Anaconda3\envs\ag10\lib\json\__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "C:\ProgramData\Anaconda3\envs\ag10\lib\json\decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "C:\ProgramData\Anaconda3\envs\ag10\lib\json\decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 192 column 411 (char 5665)

The above exception was the direct cause of the following exception:

  File "C:\ProgramData\Anaconda3\envs\ag10\lib\site-packages\ploomber\tasks\abc.py", line 554, in _build
    res = self._run()
  File "C:\ProgramData\Anaconda3\envs\ag10\lib\site-packages\ploomber\tasks\abc.py", line 663, in _run
    self.run()
  File "C:\ProgramData\Anaconda3\envs\ag10\lib\site-packages\ploomber\tasks\notebook.py", line 833, in run
    self._converter.convert()
  File "C:\ProgramData\Anaconda3\envs\ag10\lib\site-packages\ploomber\tasks\notebook.py", line 199, in convert
    self._from_ipynb(self._path_to_output, self._exporter,
  File "C:\ProgramData\Anaconda3\envs\ag10\lib\site-packages\ploomber\tasks\notebook.py", line 254, in _from_ipynb
    nb = nbformat.reads(path.read_text(), as_version=nbformat.NO_CONVERT)
  File "C:\ProgramData\Anaconda3\envs\ag10\lib\site-packages\nbformat\__init__.py", line 88, in reads
    nb = reader.reads(s, **kwargs)
  File "C:\ProgramData\Anaconda3\envs\ag10\lib\site-packages\nbformat\reader.py", line 72, in reads
    nb_dict = parse_json(s, **kwargs)
  File "C:\ProgramData\Anaconda3\envs\ag10\lib\site-packages\nbformat\reader.py", line 21, in parse_json
    raise NotJSONError(("Notebook does not appear to be JSON: %r" % s)[:77] + "...") from e
nbformat.reader.NotJSONError: Notebook does not appear to be JSON: '{\n "cells": [\n  {\n   "cell_type": "m...

ploomber.exceptions.TaskBuildError: Error building task "fit"
Found that there are related issues at github.`

Changing nb: output/nb.html to nb: output/nb.ipynb , it worked fine for me.

Wxl19980214:
## Describe your changes
Change from read_text() to read_bytes(). It looks like that read_text() will make a perfect valid json in valid by messing up its format. However, I think it might have something to do with user's language setting. I can recreate this error with my own pc which is in Chinese, but when I was trying to do it in a English based machine, this problem did not occur. 

Nevertheless changing it to read_bytes() should work the same and avoid this error because nbformat.read() can take both str and bytes as input.

Also add something in gitignore since I found out it will generate some temp files when running pytest test_notebook.py. 

Closes #870 


edublancas:
thanks for digging into this!

I'm surprised `read_bytes` works; but you're right. I did some research and `nbformat.reads`:

https://github.com/jupyter/nbformat/blob/main/nbformat/__init__.py

calls this other function:

https://github.com/jupyter/nbformat/blob/fb2b9469c5790aa12afc9d6699c66241f030a5e5/nbformat/reader.py#L45

which internally uses `json.loads` to read the notebook. `json.loads` docs say:

> Docstring:
Deserialize ``s`` (a ``str``, ``bytes`` or ``bytearray`` instance
containing a JSON document) to a Python object.

My guess is that when passed bytes, `json.loads` internally transforms the bytes into a string using the default encoding. we have run into issues in the past with encoding problems in some windows systems that don't have utf-8 as default.

and based on this (https://github.com/jupyter/nbformat/issues/181) it looks like notebooks are usually utf-8.

can you try doing `Path(...).read_text(encoding='utf-8')` and see if that fixes the issue? I think that's a better option than using bytes.

Wxl19980214:
OK will do. But I do not have access to my own pc at this moment. The pc  I am using now is English based. So I might have to wait until tmr to test if Path(...).read_text(encoding='utf-8') works.

Maybe we can communicate with the user who came up with this issue to see if his pc is in other coding instead of utf-8. That would validate our inference

edublancas:
that's fine, we can wait until tomorrow

Wxl19980214:
@edublancas  Just tested. It works with the encoding enforcement.

edublancas:
awesome!

idomic:
Nice job @Wxl19980214 - let's see if it persists now.

PR #876, opened 2022-06-23
deprecate python 3.6 (#819)

### LINKED ISSUE #819: python 3.6 end of life
https://endoflife.date/python

Python 3.6 ended one year ago, after adding 3.10 we should start thinking of deprecating 3.6.

grnnja:
## Describe your changes

updated documentation to reflect python 3.6 not being supported
removed workarounds for python 3.6

Closes #819

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Ploomber is the fastest way to build data pipelines ⚡️. Use your favorite ed
 
 ## Installation
 
-*Compatible with Python 3.6 and higher.*
+*Compatible with Python 3.7 and higher.*
 
 Install with `pip`:
 
--- a/doc/user-guide/faq/plot.rst
+++ b/doc/user-guide/faq/plot.rst
@@ -21,7 +21,7 @@ The simplest way to install ``pygraphviz`` is to use ``conda``, but you can also
 
 
 .. important::
-    If you're running Python ``3.7.x`` or lower, run: ``conda install 'pygraphviz<1.8' -c conda-forge``
+    If you're running Python ``3.7.x``, run: ``conda install 'pygraphviz<1.8' -c conda-forge``
 
 ``pip``
 *******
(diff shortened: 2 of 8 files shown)


edublancas:
Thanks for working on this and for catching all the references to Python 3.6.

I see that you removed some of the code to support 3.6. I'm unsure about this. I think we may want to keep it in case people are still using Python 3.6 but our policy should be that we won't add any more compatibility glue code in the future.

But maybe I'm overthinking it and we should get rid of this.

@idomic, I'd like to hear what you think.

grnnja:
I was thinking that since there were a few references to python 3.6 (and also a few to 3.7 and 3.8) it is inevitable that in the future that we will add code that breaks compatibility with 3.6

This would mean ploomber wouldn't work at all for 3.6 but we would still be carrying around the legacy code to support it which makes the code a bit harder to understand

Since users can use old versions of ploomber from before dropping support for 3.6, the advantage of keeping the workarounds would be they get new updates while ploomber still works for 3.6.

Essentially we may be giving a minority of users updates for a few months or a year until we add code that uses python 3.7+ features but keeping the workarounds for 3.6 in the codebase potentially forever. So I felt it was better to get rid of the code now.

edublancas:
ok, makes sense. let's drop the code then. please check the CI tests as some of them are failing

Wxl19980214:
@edublancas Can you comment this to run the integration tests?
/ok-to-test sha=eff9bc0

idomic:
@edublancas fair, we'll keep it pending until the release is done.
@grnnja any other repos that are pending on this besides this one? If not we can tick as done and move on, we'll take it from here.

grnnja:
@idomic Eduardo merged the pr for soorgeon that I opened, so it's just this one left.

idomic:
Got it!
Nothing on soopervisor/projects?

grnnja:
Oh shoot I didn't check those. I'll do that now.

grnnja:
This is what i found for soopervisor, not sure if it's worth removing though: https://github.com/ploomber/soopervisor/pull/92

Nothing for projects

idomic:
Good let's keep this one open and move on

idomic:
@edublancas did we put the notice in the latest release?
let's give it until mid Aug before merging this?

edublancas:
I haven't. created an issue https://github.com/ploomber/ploomber/issues/926

idomic (approved):
merging after new release

PR #883, opened 2022-06-27
Issue 847: CI intermittent failure on _edit_source_code

### LINKED ISSUE #847: CI fails due to an intermitent test
`tests/products/test_file.py::test_task_with_client_and_metaproduct_isnt_outdated_rtrns_waiting_download[_delete_metadata]`

https://github.com/ploomber/ploomber/runs/6765722111?check_suite_focus=true

neelasha23:
## Describe your changes
Made some modifications to `_edit_source_code` to write to a temp file and then replace final json file.

Closes #847 


tests/products/test_file.py
```diff
@@ -301,25 +301,28 @@ def test_task_without_client_is_outdated_by_code(tmp_directory):
...
     m = _load_json(path)
     m['stored_source_code'] = m['stored_source_code'] + '1 + 1\n'
-    _write_json(m, path)
+    with open(tmp_path, 'w') as f:
+        json.dump(m, f)
+    os.replace(tmp_path, path)
 
 
-def _delete_metadata(path):
+def _delete_metadata(path, tmp_path):
```
edublancas, line 313: doesn't look like tmp_path is used here

tests/products/test_file.py
```diff
@@ -301,25 +301,28 @@ def test_task_without_client_is_outdated_by_code(tmp_directory):
     assert t.product._is_outdated()
 
 
-def _edit_source_code(path):
+def _edit_source_code(path, tmp_path):
```
edublancas, line 304: why do we need this tmp_path argument?

edublancas (changes requested):
did we confirm that using os.remove fixes the problem? If so, please add a comment explaining that os.remove is preferred over the other function

tests/products/test_file.py
neelasha23: Was trying to write the json to a temporary path and then replace it with the original filename. That seemed to solve the issue that was coming up earlier.

tests/products/test_file.py
neelasha23: Ya but I had to pass it because we have a common `operation` call where `operation` can be `_delete` or `_edit_source_code`. The tmp_path is being required in `_edit_source_code`

neelasha23:
> did we confirm that using os.remove fixes the problem? If so, please add a comment explaining that os.remove is preferred over the other function

No I tried to run the CI multiple times but now the error on `_delete` is not coming up so not sure if remove solved the issue. Now the error is coming up at `_edit_source_code` and the current PR changes worked on the latest CI atleast. This is happening on Mac only

edublancas (changes requested):

neelasha23:
Added the suggested comments

edublancas:
awesome! great work!

PR #911, opened 2022-07-12
remove xfail test and add to ci integration test

### LINKED ISSUE #891: `test_cloud.py` has a lot of xskip and xfail tests
`test_cloud.py` has a lot of `xskip` and `xfail` tests. I checked the  git history and realized I added those flags since the tests were timing out. I also had to add the flags to two new tests since the backend is broken. These tests are unreliable and fail very often. I'm unsure what the solution is here but we've been trying out fixes without success (we merge to master and errors appear again). Also, I think these should be moved to the integration tests suite.

Wxl19980214:
## Describe your changes
Remove xfail tests and change test_cloud.py from ci unit to ci integration 

Closes #891 

### DIFF
--- a/.github/workflows/ci-integration-linux.yml
+++ b/.github/workflows/ci-integration-linux.yml
@@ -105,7 +105,8 @@ jobs:
         pytest src/ploomber --doctest-modules --ignore src/ploomber/resources
         unset PY_IGNORE_IMPORTMISMATCH
 
-        pytest tests/cli/test_cloud_execution.py --cov=ploomber 
+        pytest tests/cli/test_cloud_execution.py --cov=ploomber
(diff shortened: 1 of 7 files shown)


idomic (approved):
lgtm

edublancas:
I remember we had an issue that some tests were intermittently failing, is this addressing that problem? @idomic

idomic:
Those were with a skip flag, the flags were removed and the test went into the integration tests, so yes

Wxl19980214:
theres still two xskip I left there because remove it will cause failing the tests. But it looks like those tests are taken care somewhere else? Check [here](https://github.com/ploomber/ploomber/blob/master/tests/cli/test_cloud.py#L454)

PR #913, opened 2022-07-13
Large file upload exception handling

94rain:
## Describe your changes
* [x] Add exception handling for GET request in get_presigned_link and POST request in upload_zipped_project
* [x] Add `runid` as an argument of `upload_zipped_project` so that it can abort the run if there are exceptions
* [x] Add a test for `api.zip_project` when the file is too large (client-side check)
* [x] Add a test for `upload_zipped_project` (directly make a POST request).

Part of ploomber/pcloud-backend#17

### DIFF
--- a/src/ploomber/cloud/api.py
+++ b/src/ploomber/cloud/api.py
@@ -317,7 +317,7 @@ def get_presigned_link(headers):
     try:
         return _requests.get(f"{HOST}/upload", headers=headers).json()
     except Exception as err:
-        return {"Error": err}
+        return {"Error occurred while retrieving presigned upload link": err}
 
 
 def upload_zipped_project(response, verbose, runid):


src/ploomber/cloud/api.py
```diff
@@ -310,15 +314,32 @@ def zip_project(force, runid, github_number, verbose, ignore_prefixes=None):
 
 @auth_header
 def get_presigned_link(headers):
-    return _requests.get(f"{HOST}/upload", headers=headers).json()
+    try:
+        return _requests.get(f"{HOST}/upload", headers=headers).json()
+    except Exception as err:
+        return {"Error": err}
```
idomic, line 320: Make the error more informative

tests/cloud/test_api.py
```diff
@@ -61,6 +61,31 @@ def test_zip_project_ignore_prefixes(tmp_directory, sample_project):
...
+         pytest.raises(BaseException) as excinfo:
+        stat.return_value.st_size = 6 * 1024 * 1024
+        api.zip_project(force=False,
+                        runid='runid',
+                        github_number='number',
+                        verbose=False)
+    assert "Your project's source code is over 5MB" in str(excinfo.value)
+
+
+def test_upload_zipped_project_errors_if_too_large(tmp_directory):
```
idomic, line 75: This will probably fail until we push the cloud update right?

idomic (changes requested):
looking good small comments, please also remove this a file

tests/cloud/test_api.py
94rain: Yes

src/ploomber/cloud/api.py
94rain: Updated

94rain:
> Not planned.

Just curious, why is this no longer planned?

PR #920, opened 2022-07-19
Fixed confusing error when passing something.html as source #646

### LINKED ISSUE #646: confusing error when passing something.html as source
```yaml
  - source: fit.html
    product:
        nb: output/nb.ipynb
        model: output/model.pickle
```

```
(ploomber)  Edu@MBP Desktop/mlx » ploomber task fit0 -f
Loading pipeline...
Error: Failed to determine task class for source 'fit.html': list indices must be integers or slices, not str.
```

The issue is because `fit.html` is interpreted as a dotted path, hence Ploomber runs `import fit` and then tries to look for a function named `html`. However, since `fit.py` is a script, it contains this:

```python
# %% tags=["parameters"]
upstream = ['join']

# %%
df = pd.read_parquet(str(upstream['join'])) # breaks here since upstream is a list
```

yafimvo:
## Describe your changes
1. "_looks_like_path" checks a path. I've added another check if the user passed just a file name (is_path is false). We check if this file has a known extension with mimetypes. If so, this should throw the appropriate error ("Invalid extension"). Otherwise, it might be a dotted path.
2. "If you meant to import a function, please rename it." added at the end of the existing error message since there's a slight chance of a false positive where the user really wants to import a function named as the file's extension. 
3. A single file name without a path added to existing test function (test_task_class_from_script_unknown_extension)

Issue number: 646
Closes #646 

- [X ] I have performed a self-review of my code
- [X ] I have added thorough tests (when necessary).

### DIFF
--- a/tests/spec/test_taskspec.py
+++ b/tests/spec/test_taskspec.py
@@ -27,6 +27,7 @@ def test_task_class_from_script(tmp_directory, source_str, expected):
 @pytest.mark.parametrize('source_str', [
     str(Path('something', 'script.md')),
     str(Path('something', 'another', 'script.json')),
+    str(Path('file.html'))
 ])
 def test_task_class_from_script_unknown_extension(tmp_directory, source_str):
     with pytest.raises(DAGSpecInitializationError) as excinfo:
--- a/tests/spec/test_taskspec.py
+++ b/tests/spec/test_taskspec.py
@@ -27,7 +27,10 @@ def test_task_class_from_script(tmp_directory, source_str, expected):
 @pytest.mark.parametrize('source_str', [
     str(Path('something', 'script.md')),
     str(Path('something', 'another', 'script.json')),
-    str(Path('file.html'))
+    str(Path('file.html')),
+    str(Path('file.json')),
+    str(Path('file.xy')),
+    str(Path('file.a'))
 ])
 def test_task_class_from_script_unknown_extension(tmp_directory, source_str):
     with pytest.raises(DAGSpecInitializationError) as excinfo:
--- a/tests/spec/test_taskspec.py
+++ b/tests/spec/test_taskspec.py
@@ -27,10 +27,9 @@ def test_task_class_from_script(tmp_directory, source_str, expected):
 @pytest.mark.parametrize('source_str', [
     str(Path('file.html')),
     str(Path('file.json')),
-    str(Path('a.a')),
-    str(Path('file.b'))
+    str(Path('file.a'))
 ])
-def test_task_class_from_file_without_path_unknown_extension(
+def test_task_class_from_valid_file_without_path_unsupported_extension(
         tmp_directory, source_str):
     with pytest.raises(DAGSpecInitializationError) as excinfo:
         task_class_from_source_str(source_str,
--- a/tests/spec/test_taskspec.py
+++ b/tests/spec/test_taskspec.py
@@ -25,12 +25,28 @@ def test_task_class_from_script(tmp_directory, source_str, expected):
 
 
 @pytest.mark.parametrize('source_str', [
-    str(Path('something', 'script.md')),
-    str(Path('something', 'another', 'script.json')),
     str(Path('file.html')),
     str(Path('file.json')),
-    str(Path('file.xy')),
-    str(Path('file.a'))
+    str(Path('a.a')),
+    str(Path('file.b'))
+])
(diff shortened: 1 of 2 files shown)


idomic:
@yafimvo Is this ready for review? 
Also please fix the ci

94rain:
Seems that `_looks_like_path` will be evaluated to true for `.a` extensions with the changes, which raises the exception and causes `tests/spec/test_dagspec.py::test_python_callables_with_extract_upstream` to fail. I'm not very familiar with this part. Should we move the `elif` to a later point?

```python
>>> import mimetypes
>>> mimetypes.guess_type('a.a')
('application/octet-stream', None)
```


https://github.com/ploomber/ploomber/blob/c00102227cfb9f8650e0bc49b6780a7b1b6f562f/src/ploomber/spec/taskspec.py#L94-L99

yafimvo:
1. Restored `_looks_like_path` to its original implementation. 
2. Removed `mimetypes.guess_type` check
3. The error message 'If you meant to import a function, please rename it.' was added to a later point (after import fails)

If it's not a path it might be a dotted path (even with a name such as 'file.a'), so we try to import it. If the import fails it might be because of an invalid extension.

tests/spec/test_taskspec.py
```diff
@@ -24,9 +24,29 @@ def test_task_class_from_script(tmp_directory, source_str, expected):
         source_str, lazy_import=False, reload=False, product=None) is expected
 
 
+@pytest.mark.parametrize('source_str', [
+    str(Path('file.html')),
+    str(Path('file.json')),
+    str(Path('a.a')),
+    str(Path('file.b'))
```
idomic, line 31: looks like this test is failing

idomic (approved):
lgtm!
Nice job

idomic:
@yafimvo please make sure it behaves as expected,
I tried running `ploomber task sample -f` and `ploomber build -f` on this pipeline.yaml:
```
tasks:
  - source: sample.html
    product: output/nb.ipynb
    name: sample
```

The expected behavior was seeing the exception, but it actually tried running the task and complaining on missing out:

```
ploomber task sample -f
Loading pipeline...
100%|███████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 747.91it/s]
  File "/Documents/ploomber/ploomber/src/ploomber/tasks/abc.py", line 583, in _build
    self._post_run_actions()
  File "/Documents/ploomber/ploomber/src/ploomber/tasks/abc.py", line 376, in _post_run_actions
    raise TaskBuildError(f'Error building task {self.name!r}: '
ploomber.exceptions.TaskBuildError: Error building task 'sample': the task ran successfully but the product File('output/nb.ipynb') is missing. Ensure your code is generating it.

The above exception was the direct cause of the following exception:

  File "/Documents/ploomber/ploomber/src/ploomber/cli/io.py", line 52, in wrapper
    fn(**kwargs)
  File "/Documents/ploomber/ploomber/src/ploomber/telemetry/telemetry.py", line 500, in wrapper
    raise e
  File "/Documents/ploomber/ploomber/src/ploomber/telemetry/telemetry.py", line 488, in wrapper
    result = func(*args, **kwargs)
  File "/Documents/ploomber/ploomber/src/ploomber/cli/task.py", line 76, in main
    _task_cli()
  File "/Documents/ploomber/ploomber/src/ploomber/cli/task.py", line 61, in _task_cli
    task.build(force=args.force)
  File "/Documents/ploomber/ploomber/src/ploomber/tasks/abc.py", line 522, in build
    res, _ = self._build(catch_exceptions=catch_exceptions)
  File "/Documents/ploomber/ploomber/src/ploomber/tasks/abc.py", line 595, in _build
    raise TaskBuildError(msg) from e
ploomber.exceptions.TaskBuildError: Exception when running on_finish for task 'sample'

ploomber.exceptions.TaskBuildError: Error building task 'sample': the task ran successfully but the product File('output/nb.ipynb') is missing. Ensure your code is generating it.
ploomber.exceptions.TaskBuildError: Exception when running on_finish for task 'sample'

```

idomic (changes requested):
Comment above

yafimvo:
I've tried to run the same commands with the same `pipeline.yaml` on my local env and couldn't reproduce this error (receive "invalid extension message")

Did you run this command on your local env manually?

idomic (approved):
After retrying looks like it works as expected!

PR #927, opened 2022-07-20
start_method added to parallel executor

### LINKED ISSUE #882: Allowing to switch start_method in Parallel executor
I have some Pytorch code which I want to run on GPU (`net.cuda()` and whatnot)

I get the following error:
```
RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method
```
I tried to fix this by either calling
```
mp.set_start_method('spawn', force=True)
```

or 
```
torch.multiprocessing.set_start_method('spawn')
```

In the first case nothing happens, and in the second I get `RuntimeError: context has already been set`

How to set multiprocessing start method with ploomber?

yafimvo:
## Describe your changes
Added a new start_method argument to Parallel.py and then initiated ProcessPoolExecutor with mp_context where context is default or the given start_method

* I can't connect to GPU locally so I need to test this code on AWS VM.

Closes #882 

### DIFF
--- a/tests/dag/test_dag.py
+++ b/tests/dag/test_dag.py
@@ -988,13 +988,21 @@ def test_early_stop_from_task_level_on_finish(executor, tmp_directory):
 # test early stop when registered an on_failure hook, maybe don't run hook?
 
 
-@pytest.mark.parametrize('start_method', ["invalid_start_method_value"])
+@pytest.mark.parametrize('start_method', ['invalid_start_method_value'])
 def test_invalid_start_method_in_parallel_executor(tmp_directory,
                                                    start_method):
     with pytest.raises(ValueError) as excinfo:
         DAG(executor=Parallel(processes=2, start_method=start_method))
 
     assert 'Invalid start_method' in str(excinfo.value)
+
+
+@pytest.mark.parametrize('start_method', ['spawn', 'fork', 'forkserver'])
+def test_switch_start_method_in_parallel_executor(tmp_directory, start_method,
+                                                  monkeypatch):
+    dag = DAG(executor=Parallel(processes=2, start_method=start_method))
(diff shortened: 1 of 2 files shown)


yafimvo:
I've successfully reproduced this exception on AWS VM running the code on GPU with Cuda.
After running the new code it seems to work OK using following pipeline.yaml configuration:
```
executor:
  dotted_path: ploomber.executors.Parallel
  processes: 4
  start_method: spawn 
```

I also added a ValueType exception when `start_method` is incorrect (possible values are: `spawn, fork, forkserver`)

I'm not sure how to test this parallel behavior.

idomic:
Nice!

> I'm not sure how to test this parallel behavior.

You'd need to run a pipeline.yaml and tasks that has independent tasks that can run in parallel.
Check out [this grid feature](https://github.com/ploomber/projects/blob/cbcf4aee763d2eac33d0d749faa3807c9b7c07d9/cookbook/grid/README.md) pipeline.

tests/dag/test_dag.py
```diff
@@ -993,6 +988,15 @@ def test_early_stop_from_task_level_on_finish(executor, tmp_directory):
 # test early stop when registered an on_failure hook, maybe don't run hook?
 
 
+@pytest.mark.parametrize('start_method', ["invalid_start_method_value"])
```
idomic, line 991: Also add a test on the other start methods and validate it's set in the context

idomic (changes requested):
Looks good small comment, and missing the manual test on parallel execution

idomic (approved):
Nice work!
For next PRs let's try starting with some tests and then develop (TDD)

PR #931, opened 2022-07-21
Migrate telemetry to ploomber_core

Wxl19980214:
## Describe your changes
Delete telemetry related code. 
Kept tests for command calls telemetry correctly. 
Import telemetry from ploomber_core 
Changed a few definition of functions so we need to pass in package name and version now.


Wxl19980214:
@idomic Please do a ok-to-test

PR #949, opened 2022-07-27
Tutorial unit test

### LINKED ISSUE #799: add unit tests for onboarding tutorial
I added a small tutorial (triggered with `python -m ploomber.onboard`) but it doesn't have tests yet

Wxl19980214:
## Describe your changes
Added a few unit tests.

However, this is a bit different from what i thought the tests should be.  I was thinking about an end2end test testing whether a correct user action would lead to the correct next step, vice versa. 

I couldn't find a way to do that. So for now, I am mocking every external calls and other functions to make sure the entire workflow runs once and check we have corresponding outputs.

Closes #799 

### DIFF
--- a/tests/onboard/test_onboard.py
+++ b/tests/onboard/test_onboard.py
@@ -9,8 +9,7 @@ from ploomber.onboard import __main__ as onboard
 
 from ploomber.cli import examples
 from ploomber import DAG
-from ploomber.cli import cloud
-
+import ploomber
 
 def test_make_open_callable():
     callable_open = onboard._make_open_callable('open')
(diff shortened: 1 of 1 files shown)


edublancas:
is this ready for review? I see it's still draft

Wxl19980214:
It is ready for review

Wxl19980214:
@edublancas  also can you retry ok-to-test, I see it not triggering the workflow for some reason

edublancas:
looks like the ok to test is broken, I'll merge since this is not affecting the integration tests

Wxl19980214:
Yeah the ok-to-test is broken for some reason, I will connect with Ido on it. Might be a key error or some sort

PR #960, opened 2022-08-01
Telemetry remove

Wxl19980214:
## Describe your changes

Removed telemetry. 

Use Telemetry as a class object to pass in version, pacakge_name, and posthog api key.

We need to update posthog api_key and put it in init.py


Wxl19980214:
I still need to put in the api key for posthog

edublancas:
hi @Wxl19980214 - I made a small change, I created a module so we create the telemetry instance there and import; instead of creating a new instance on each module. this will fail since I moved a few things around, please fix all the broken instances, check what I did here: https://github.com/ploomber/ploomber/pull/960/commits/2552107fc978018e4dddc7a2fe8862c3f8791f5c

idomic:
@Wxl19980214 please patch a fix.

Wxl19980214:
fixed. ready for review

edublancas:
nice work!

PR #962, opened 2022-08-02
copy when click on title of terminal in docs

grnnja:
## Describe your changes

Added the title of the terminal to the clickable area for copying terminal text in the docs.

Closes https://github.com/ploomber/projects/issues/51

### DIFF
--- a/doc/_static/js/custom.js
+++ b/doc/_static/js/custom.js
@@ -1,3 +1,8 @@
+// https://stackoverflow.com/questions/31128855/comparing-ecma6-sets-for-equality
+const eqSet = (xs, ys) => 
+    xs.size === ys.size &&
+    [...xs].every((x) => ys.has(x));
+
 $(document).ready(function () {
     addTerminalStyle(
         $('[class="highlight-bash notranslate"]'),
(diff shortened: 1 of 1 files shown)


doc/_static/js/custom.js
```diff
@@ -26,17 +26,26 @@ $(document).ready(function () {
     // TODO: this should listen for click events in the whole terminal
     // window, not only in the content sub-window
     // Clicking on the terminnals copies the content to the clipboard
-    $("div.highlight").click(function () {
-        if ("jquery: ", $(this).parent().attr('class') == "doctest highlight-default notranslate") {
-            let text = $(this).text();
+    $("[class*='highlight-']") .click(function () {
+        var formatCopiedText = function(text) {
```
idomic, line 30: please change to let, var usage isn't recommended after ES 5 or 6 I think

idomic (changes requested):
please fix

idomic:
@yafimvo can you please review as well?

doc/_static/js/custom.js
grnnja: fixed

yafimvo (commented):
`if ("jquery: ", $(this).attr('class') == "doctest highlight-default notranslate")`

1. When type coercion is not required, it is recommended to use '===' instead of '=='
2. It may not be necessary to use the comma operator here
3. Elements' class names may change in the future so this `doctest highlight-default notranslate` string format may be different and the comparison will fail. You may avoid this by comparing substrings.

grnnja:
Good points I will fix this. Though it is kinda hard to tell from the diff I didn't actually write this code. I moved some other code but it shows in the diff that the `if` was moved. @idomic am I typically responsible for other people's code?

idomic:
@grnnja I'm not sure how it got there, but let's take the opportunity to fix that :)
It's just a better practice to improve bit by bit and overall in the long run (instead of copy pasting, cleaning a bit etc).

idomic:
luckily that's the only place we're using jquery we can simply avoid using it 👍

doc/_static/js/custom.js
```diff
@@ -26,17 +26,30 @@ $(document).ready(function () {
...
+            return text;
         }
+
+        // https://stackoverflow.com/questions/31128855/comparing-ecma6-sets-for-equality
+        const eqSet = (xs, ys) => 
+            xs.size === ys.size &&
+            [...xs].every((x) => ys.has(x));
+
+        const terminalText = $(this).find('div.highlight').text();
+        const isPythonConsole = eqSet(new Set($(this).attr('class').split(' ')),
```
yafimvo, line 44: isPythonConsole - Are you sure this comparison is ever true? The term 'doctest' didn't turn up anything on my search.

yafimvo (changes requested):

doc/_static/js/custom.js
```diff
@@ -26,17 +26,30 @@ $(document).ready(function () {
...
             text = textArr.join("\n");
             text = text.replaceAll("... ", "").replaceAll(">>> ", "");
-            navigator.clipboard.writeText(text);
-        } else {
-            navigator.clipboard.writeText($(this).text());
+            return text;
         }
+
+        // https://stackoverflow.com/questions/31128855/comparing-ecma6-sets-for-equality
+        const eqSet = (xs, ys) =>
```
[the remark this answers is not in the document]
grnnja: fixed!

doc/_static/js/custom.js
grnnja: Yeah, like for this page, the terminals have the classes: https://docs.ploomber.io/en/latest/api/_modules/root/ploomber.DAG.html#ploomber.DAG

I also couldn't find it in the repo so I'd assume it's generated by sphinx

yafimvo (approved):

idomic:
Good job!

PR #969, opened 2022-08-05
improve readme getting started

### LINKED ISSUE #940: readme improvements
Our README has too many options and people get lost. It should be clear how to get started.

I think it should only give two options: running the onboarding tutorial locally or taking them to the first tutorial.

and we should replace this with a larger button since people don't see it:

image

grnnja:
Closes #940


idomic:
@edblancas thoughts on the link? Maybe a button will be more visible. Also we might want to do something with the top binder button.

edublancas:
yeah, let's add a button. You can copy this and manually edit it to change the label. put something like "Open in JupyterLab" https://github.com/ploomber/ploomber/blob/master/_static/get-started.svg

I think we can get rid of the binder button

grnnja:
Updated!

idomic (approved):
lgtm, nice job!

PR #971, opened 2022-08-08
Remove parameters cell missing exception and add cell when it is missing instead

### LINKED ISSUE #950: do not require a parameters cell in the Python API
Currently, we throw an error if the notebooks are missing a parameters cell, but this is too restrictive; we should show a warning instead.

I think this does don't happen with the spec API and we add the parameters cell, but I must check.

94rain:
## Describe your changes
* Change exception to warnings for missing cells in `_post_init_validation`
* Update exception tests `test_error_if_parameters_cell_doesnt_exist` and `test_error_missing_params_cell_shows_path_if_available` to check warnings instead
* Create a method `_validate_parameters_cell` in the class NotebookMixin in `src/ploomber/tasks/notebook.py` to check and add parameter cells if missing

Closes #950

### DIFF
--- a/src/ploomber/sources/notebooksource.py
+++ b/src/ploomber/sources/notebooksource.py
@@ -40,8 +40,7 @@ from jupytext.formats import long_form_one_format, short_form_one_format
 from jupytext.config import JupytextConfiguration
 import parso
 
-from ploomber.exceptions import (SourceInitializationError,
-                                 MissingParametersCellError)
+from ploomber.exceptions import SourceInitializationError
 from ploomber.placeholders.placeholder import Placeholder
 from ploomber.util import requires
 from ploomber.sources.abc import Source
@@ -403,7 +402,7 @@ class NotebookSource(Source):
                                             'parameters')
 
         if params_cell is None:
-            loc = ' "{}"'.format(self.loc) if self.loc else ''
+            loc = ' "{}"'.format(pretty_print.try_relative_path(self.loc)) if self.loc else ''
             msg = ('Notebook{} does not have a cell tagged '
                    '"parameters"'.format(loc))
 
@@ -421,8 +420,7 @@ Go to: https://ploomber.io/s/params for more information
                 msg += ('. Add a cell at the top and tag it as "parameters". '
                         'Go to the next URL for '
                         'details: https://ploomber.io/s/params')
-
-            raise MissingParametersCellError(msg)
+            click.secho(msg)
 
     def _post_render_validation(self):
         """
@@ -544,7 +542,7 @@ Go to: https://ploomber.io/s/params for more information
     def _get_parameters_cell(self, force=False):
         self._read_nb_str_unrendered(force=force)
         cell, _ = find_cell_with_tag(self._nb_obj_unrendered, tag='parameters')
-        return cell.source
+        return cell.source if cell else None
 
     def extract_upstream(self, force=False):
         extractor_class = extractor_class_for_language(self.language)
@@ -1038,7 +1036,7 @@ def _warn_on_unused_params(nb, params):
                       f'task\'s source code: {pretty_print.iterable(unused)}')
 
 
-def add_parameters_cell(path, extract_upstream, extract_product):
+def add_parameters_cell(path, extract_upstream=False, extract_product=False):
     """
     Add parameters cell to a script/notebook in the given path, overwrites the
     original file
--- a/src/ploomber/sources/notebooksource.py
+++ b/src/ploomber/sources/notebooksource.py
@@ -401,11 +401,13 @@ class NotebookSource(Source):
                                   extract_product=False):
         '''Check parameters call and add it when it's missing
 
-        Keyword arguments:
-        extract_upstream -- Flags used to determine the content of
-        the parameters cell, only used if the notebook is
-        missing the parameters cell (default False)
-        extract_product -- Same as extract_upstream (default False)
+        Parameters
+        ----------
+        extract_upstream : bool, default: True
+            Flags used to determine the content of the parameters cell,
+            only used if the notebook is missing the parameters cell
+        extract_product : bool, default: True
+            Same as extract_upstream (default False)
         '''
         params_cell, _ = find_cell_with_tag(self._nb_obj_unrendered,
                                             'parameters')
@@ -1053,10 +1055,7 @@ product = None
 """
 
     if not extract_upstream and not extract_product:
-        source += """
-upstream = None
-product = None
-"""
+        source += '# add default values for parameters here'
 
     c = JupytextConfiguration()
     c.notebook_metadata_filter
--- a/src/ploomber/sources/notebooksource.py
+++ b/src/ploomber/sources/notebooksource.py
@@ -402,7 +402,8 @@ class NotebookSource(Source):
                                             'parameters')
 
         if params_cell is None:
-            loc = ' "{}"'.format(pretty_print.try_relative_path(self.loc)) if self.loc else ''
+            loc = ' "{}"'.format(pretty_print.try_relative_path(self.loc)) \
+                if self.loc else ''
             msg = ('Notebook{} does not have a cell tagged '
                    '"parameters"'.format(loc))
 
--- a/src/ploomber/sources/notebooksource.py
+++ b/src/ploomber/sources/notebooksource.py
@@ -1020,7 +1020,7 @@ def _warn_on_unused_params(nb, params):
     _, idx = find_cell_with_tag(nb, 'parameters')
 
     # the notebooks might not have a parameters cell
-    if not idx:
+    if idx is None:
         return
 
     del nb.cells[idx]
--- a/src/ploomber/sources/notebooksource.py
+++ b/src/ploomber/sources/notebooksource.py
@@ -423,6 +423,30 @@ Go to: https://ploomber.io/s/params for more information
                         'details: https://ploomber.io/s/params')
             click.secho(msg)
 
+    def _validate_parameters_cell(self,
+                                  extract_upstream=False,
+                                  extract_product=False):
+        '''Check parameters call and add it when it's missing
+
+        Keyword arguments:
+        extract_upstream -- Flags used to determine the content of
+        the parameters cell, only used if the notebook is
+        missing the parameters cell (default False)
+        extract_product -- Same as extract_upstream (default False)
+        '''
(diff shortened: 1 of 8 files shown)


src/ploomber/sources/notebooksource.py
```diff
@@ -419,8 +419,7 @@ def _post_init_validation(self, value):
                 msg += ('. Add a cell at the top and tag it as "parameters". '
                         'Go to the next URL for '
                         'details: https://ploomber.io/s/params')
-
-            raise MissingParametersCellError(msg)
+            click.secho(msg)
```
edublancas, line 422: since we are adding the cell anyway, we no longer need this message

src/ploomber/tasks/notebook.py
```diff
@@ -268,9 +271,28 @@ def _from_ipynb(path_to_nb, exporter, nbconvert_export_kwargs):
...
+    def _validate_parameters_cell(source,
+                                  extract_up=False,
+                                  extract_prod=False):
+        params_cell, _ = find_cell_with_tag(source._nb_obj_unrendered,
+                                            'parameters')
+
+        if params_cell is None:
+            loc = pretty_print.try_relative_path(source.loc)
+            # raise MissingParametersCellError(msg)
+            # import pdb; pdb.set_trace()
```
edublancas, line 285: remove this commented lines

src/ploomber/tasks/notebook.py
```diff
@@ -669,7 +691,10 @@ def __init__(self,
                  nbconvert_export_kwargs=None,
                  local_execution=False,
                  check_if_kernel_installed=True,
-                 debug_mode=None):
+                 debug_mode=None,
+                 extract_up=False,
+                 extract_prod=False
```
edublancas, line 696: we dont need this in the class constructor right? these will always be False when initializing the class

src/ploomber/tasks/notebook.py
```diff
@@ -697,6 +722,8 @@ def __init__(self,
             kernelspec_name,
             static_analysis,
             check_if_kernel_installed,
+            extract_up,
+            extract_prod
```
edublancas, line 726: we can pass False to both here

src/ploomber/tasks/notebook.py
```diff
@@ -757,7 +784,18 @@ def _init_source(source,
...
+        ns = NotebookSource(
+            source,
+            ext_in=ext_in,
+            kernelspec_name=kernelspec_name,
+            static_analysis=static_analysis,
+            check_if_kernel_installed=check_if_kernel_installed,
+            **kwargs)
+        NotebookMixin._validate_parameters_cell(ns, extract_up, extract_prod)
+        # re-create NotebookSource instance
         return NotebookSource(
```
edublancas, line 799: do we need to initialize this two times? why not

```
ns  = NotebookSource(...)
ns._validate_parameters_cell(extract_up, extract_prod)
return ns
```

src/ploomber/tasks/notebook.py
```diff
@@ -910,17 +948,31 @@ def __init__(self,
                  params=None,
                  ext_in=None,
                  static_analysis='regular',
-                 local_execution=False):
+                 local_execution=False,
+                 extract_up=False,
+                 extract_prod=False
```
edublancas, line 953: same as NotebookRunner, extract_up and extract_prod will always be False so no need to have them here

src/ploomber/tasks/notebook.py
```diff
@@ -910,17 +948,31 @@ def __init__(self,
...
+                 ):
         self.ext_in = ext_in
         self.local_execution = local_execution
 
         kwargs = dict(hot_reload=dag._params.hot_reload)
         self._source = ScriptRunner._init_source(source, kwargs, ext_in,
-                                                 static_analysis)
+                                                 static_analysis, extract_up,
+                                                 extract_prod)
```
edublancas, line 961: extract_up and extract_prod can be False here

src/ploomber/tasks/notebook.py
```diff
@@ -910,17 +948,31 @@ def __init__(self,
...
+                     extract_up=False, extract_prod=False):
+        ns = NotebookSource(
+            source,
+            ext_in=ext_in,
+            static_analysis=static_analysis,
+            kernelspec_name=None,
+            check_if_kernel_installed=False,
+            **kwargs,
+        )
+        NotebookMixin._validate_parameters_cell(ns, extract_up, extract_prod)
```
edublancas, line 975: same as with NotebookRunner, do we need to initialize two times?

src/ploomber/tasks/notebook.py
```diff
@@ -268,9 +271,28 @@ def _from_ipynb(path_to_nb, exporter, nbconvert_export_kwargs):
 
 
 class NotebookMixin(FileLoaderMixin):
+
+    @staticmethod
+    def _validate_parameters_cell(source,
+                                  extract_up=False,
+                                  extract_prod=False):
```
edublancas, line 278: please document extract_up and extract_prod, also rename them for clarity to extract_upstream and extract_product

description: flags used to determine the content of the parameters cell, only used if the notebook is missing the parameters cell

src/ploomber/sources/notebooksource.py
```diff
@@ -1036,7 +1035,7 @@ def _warn_on_unused_params(nb, params):
                       f'task\'s source code: {pretty_print.iterable(unused)}')
 
 
-def add_parameters_cell(path, extract_upstream, extract_product):
+def add_parameters_cell(path, extract_upstream=False, extract_product=False):
```
edublancas, line 1038: add a new confition here

```
if not extract_upstream and not extract_product:
    source += '# add default values for parameters here'
```

edublancas (changes requested):

src/ploomber/sources/notebooksource.py
94rain: Do we need this? Looks like they might be redundant.
```python
    if extract_upstream:
        source += """\
# declare a list tasks whose products you want to use as inputs
upstream = None
"""
    if extract_product:
        source += """\
# declare a dictionary with the outputs of this task
product = None
"""
```

src/ploomber/sources/notebooksource.py
94rain: Do we want to remove this whole method (`_post_init_validation`)?

src/ploomber/tasks/notebook.py
94rain: Working on it

94rain:
All except the two open conversations above have been fixed.

src/ploomber/sources/notebooksource.py
edublancas: keep the method but leave it empty (add a `pass` to the body), as other parts of the framework call it

src/ploomber/sources/notebooksource.py
edublancas: why redudnant? the first two apply when any of them is True or both. this new one applies when none of them are true

src/ploomber/sources/notebooksource.py
```diff
@@ -393,34 +392,31 @@ def _post_init_validation(self, value):
...
+                                  extract_upstream=False,
+                                  extract_product=False):
+        '''Check parameters call and add it when it's missing
+
+        Keyword arguments:
+        extract_upstream -- Flags used to determine the content of
+        the parameters cell, only used if the notebook is
+        missing the parameters cell (default False)
+        extract_product -- Same as extract_upstream (default False)
+        '''
```
edublancas, line 407: please use numpydoc format for the docstring https://numpydoc.readthedocs.io/en/latest/format.html

src/ploomber/sources/notebooksource.py
```diff
@@ -1052,6 +1048,12 @@ def add_parameters_cell(path, extract_upstream, extract_product):
         source += """\
 # declare a dictionary with the outputs of this task
 product = None
+"""
+
+    if not extract_upstream and not extract_product:
+        source += """
+upstream = None
+product = None
```
edublancas, line 1056: the sample content should be a comment:

```
source += '# add default values for parameters here'
```

tests/sources/test_notebooksource.py
```diff
@@ -256,25 +256,6 @@ def test_error_if_source_is_dir_suggest_scaffold(tmp_directory):
...
-    path = Path('nb.ipynb')
-    path.write_text(new_nb(fmt='ipynb', add_tag=False))
-
-    with pytest.raises(SourceInitializationError) as excinfo:
-        NotebookSource(path)
-
-    assert ('Notebook "nb.ipynb" does not have a cell tagged "parameters"'
-            in str(excinfo.value))
-
-
```
edublancas, line 277: Do we have tests to check that the parameter cell is added if the input notebook *does not* have it? I couldn't see them

you can use this as reference. read the source ipynb file with nbformat and check that the top cell has the parameters tag

edublancas (changes requested):

tests/sources/test_notebooksource.py
94rain: I think we already have `tests/sources/test_notebooksource.py::test_add_parameters_cell` in place. Would that be sufficient?

src/ploomber/sources/notebooksource.py
94rain: Updated:
        '''Check parameters call and add it when it's missing

        Parameters
        ----------
        extract_upstream : bool, default: True
            Flags used to determine the content of the parameters cell,
            only used if the notebook is missing the parameters cell
        extract_product : bool, default: True
            Same as extract_upstream (default False)
        '''

src/ploomber/sources/notebooksource.py
94rain: Sorry, I misunderstood that. Fixed.

tests/sources/test_notebooksource.py
94rain: Added in https://github.com/ploomber/ploomber/pull/971/commits/2c624eebce91c6953ee366a2ab0d730670efe14d

94rain:
Updated the CHANGELOG as well (https://github.com/ploomber/ploomber/pull/971/commits/133c10debffc798aaddac5e824dbf83fd17256d2)

edublancas:
great, I did a minor fix because I found an edge case: the pipeline would break if the input had no cells

I'll just wait for the CI and merge but it should be good now since it passed already

edublancas (approved):

Issue #797, opened 2022-05-19
inject cell error after pairing notebooks


edublancas:
when a script is in a subdirectory (say `scripts/fit.py`) and we pair notebooks in a certain folder (say `ploomber nb --pair notebooks`), the notebooks are created relative to the scripts (`scripts/notebooks/fit.ipynb`), and the metadata is stored using a relative path (`notebooks`)

However, if we run the `ploomber nb --inject` command, ploomber will read the relative path in the metadata `notebooks` and it will fail since it won't find the notebooks (stored in `scripts/notebooks`)

The solution is to read the paired notebooks relative to the scripts.

To reproduce:

```
ploomber examples -n templates/ml-basic -o ml
# move fit.py to scripts/fit.py and update the pipeline.yaml

ploomber nb --pair notebooks

ploomber nb --inject
```

edublancas:
this has been fixed

PR #979, opened 2022-08-17
Bug#797

amdflake:
Closes bug #797


tests/sources/test_notebooksource.py
```diff
@@ -636,6 +636,103 @@ def test_save_injected_cell_in_paired_notebooks(tmp_nbs, prefix):
     assert get_injected_cell(jupytext.read(Path(prefix, 'load.ipynb')))
     assert get_injected_cell(jupytext.read(Path('load.py')))
 
+def test_move_script_save_injected_cell_in_paired_notebooks(tmp_nbs):
```
idomic, line 639: test_move_script_save_injected_cell_in_paired_notebooks(tmp_nbs, injection_setup)

fixture: def injection_setup:
```
Path('scripts').mkdir(parents=True, exist_ok=True)
Path('load.py').rename('scripts/load.py')
Path('clean.py').rename('scripts/clean.py')
Path('plot.py').rename('scripts/plot.py')
```

idomic: https://docs.pytest.org/en/6.2.x/fixture.html

tests/sources/test_notebooksource.py
```diff
@@ -636,6 +636,103 @@ def test_save_injected_cell_in_paired_notebooks(tmp_nbs, prefix):
...
+    Path('scripts').mkdir(parents=True, exist_ok=True)
+    Path('load.py').rename('scripts/load.py')
+    Path('clean.py').rename('scripts/clean.py')
+    Path('plot.py').rename('scripts/plot.py')
+
+    pipeline = """meta:
+  extract_upstream: False
+  extract_product: False
+
+tasks:
```
idomic, line 650: Use the template pipeline

tests/sources/test_notebooksource.py
idomic: conftest

tests/assets/nbs-moved-scripts/pipeline.yaml
```diff
@@ -0,0 +1,25 @@
+meta:
```
idomic, line 1: @edublancas can we use instead of this new asset in one of the existing one and inject on top of it?

edublancas: just to confirm: looks like this `nbs-moved-scripts` was created to represent the user case where the `source`s are nested, right? if that's the only reason, we can use `nb-nested` to avoid creating a new assets folder just for this test.

tests/sources/test_notebooksource.py
amdflake: @idomic If I use `nb-nested` should I still make this fixture? Since the files are already in nested folders, I wasn't sure if it was necessary.

amdflake: Please disregard that last message, I didn't realize this was a comment from the first commit.

tests/assets/nbs-moved-scripts/pipeline.yaml
amdflake: Just changed the test so that it uses `nb-nested` instead!

idomic: Perfect, I think if you already have the fixture and template, there's no need to reinvent the wheel.

PR #981, opened 2022-08-17
Add priority inject function

### LINKED ISSUE #902: Injecting parameters when the same notebook appears more than once

```
ploomber nb --inject
```

```yaml
tasks:
  - source: template.ipynb
     name: task-a
     product: report-a.ipynb
     params:
       some_param: a

  - source: template.ipynb
     name: task-b
     product: report-b.ipynb
     params:
       some_param: b
```

If we execute:

```
ploomber nb --inject
```

Example:

```
```

```
```

```
```

Wxl19980214:
## Describe your changes
After turning on priority injection option in setup.cfg, if multiple tasks share the same notebook as source, only content of the task name ends with -suffix will be injected to the source notebook.

## Concerns
From my implementation the wildcard "*" in "*-suffix" looks like unrelated. As long as user put inject-priority = anything here, I treated it as turning on this functionality. But indeed the task name has to strictly follow the naming convention, so I think we need to properly document it.

For now it works with injection parameters, I am not sure about the dev env where there can have multiple output? In (here)[https://github.com/ploomber/ploomber/issues/902#issuecomment-1206843741] looks like we have multiple copies in execution of dev env, how should I reproduce this situation?    

Closes #902

- [] I have added the right documentation (when needed). Product update? If yes, write one line about this update.

### DIFF
--- a/src/ploomber/cli/nb.py
+++ b/src/ploomber/cli/nb.py
@@ -2,6 +2,7 @@ import argparse
 import json
 import shutil
 from pathlib import Path
+from collections import defaultdict
 import stat
 
 import click
@@ -11,6 +12,8 @@ from ploomber.cli.io import command_endpoint
 from ploomber.sources.notebooksource import recursive_update
 from ploomber_core.exceptions import BaseException
 from ploomber.telemetry import telemetry
+from ploomber.util import config
+from ploomber.util.default import find_file_recursively
 
 
 def _format(fmt, entry_point, dag, verbose=True):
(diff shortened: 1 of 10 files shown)


yafimvo:
Based on the original ticket discussion and after reviewing the PR there are a few things I'd like to point out

1. It doesn't inject parameters into the original template file
	- Should we inject the prioritized params to the original template.ipynb?
2. Creates a version of each template with injected params regardless of the given `inject-priority` in `setup.cfg`
	- When `inject-priority` is given do we want to create copies of template.ipynb with non prioritized parameters?
	- What to expect when `inject-priority` is not given?
3. Warnings and errors are not implemented
4. Missing tests

PR #985, opened 2022-08-21
902 cli cell injection with priority

### LINKED ISSUE #902: Injecting parameters when the same notebook appears more than once

yafimvo:
## Describe your changes
Implemented the CLI approach that allows to choose which set of parameters to inject into the notebook from `pipeline.yaml`

`ploomber nb --inject --priority task-a`
`ploomber nb --inject --priority task-a --priority task-c`

Closes #902 

- [X] I have added thorough tests (when necessary).


CHANGELOG.md
```diff
@@ -1,9 +1,10 @@
 # CHANGELOG
 
-## 0.21.1dev
```
idomic, line 3: @edblancas what's the approach here, are we keeping the `dev` version for dev string updates?

tests/cli/test_nb.py
```diff
@@ -24,6 +24,217 @@ def git_commit():
...
+
+def test_inject_invalid_prioritized_task_single_task(monkeypatch, capsys,
+                                                     path_to_assets):
+    # ploomber nb --inject --priority this-task-doesnt-exist
+    nb_inject_assets_path = f'{path_to_assets}/test-nb-inject-assets'
+    template_path = f'{nb_inject_assets_path}/template.ipynb'
+    test_pipeline = f'{nb_inject_assets_path}/pipeline.yaml'
+
+    Path(test_pipeline).write_text(f"""
+tasks:
```
idomic, line 175: This and the 3 lines below:

```
nb_inject_assets_path = f'{path_to_assets}/test-nb-inject-assets'
template_path = f'{nb_inject_assets_path}/template.ipynb'
test_pipeline = f'{nb_inject_assets_path}/pipeline.yaml'
```

Are repetitive, let's export it to a function and reuse the code.

tests/assets/test-nb-inject-assets/template_b.ipynb
```diff
@@ -0,0 +1,86 @@
+{
```
idomic, line 1: Is there an existing pipeline we can leverage here?

src/ploomber/dag/util.py
```diff
@@ -11,6 +11,17 @@
 from ploomber.io import pretty_print
 
 
+def get_unique_list(list):
+    unique = []
+    duplicates = []
+    for item in list:
```
idomic, line 17: Maybe use set()?

src/ploomber/cli/nb.py
```diff
@@ -11,6 +11,94 @@
...
+                                             prioritized_task_name_to_inject)
+        is_prioritized_task_exist = True if dag_task is not None else False
+        if is_prioritized_task_exist:
+            full_template_name = Path(dag_task.source._path).name
+            if full_template_name not in used_tempaltes:
+                used_tempaltes.append(full_template_name)
+            else:
+                exceptions.append(
+                    BaseException(
+                        'Error. Values are correspond to the same task.'))
```
idomic, line 80: typo?
`Values correspond`

idomic (changes requested):
Added few comments besides that looks good.

tests/assets/test-nb-inject-assets/template_b.ipynb
yafimvo: I think there are some pipelines we can use but it might be confusing to debug/create more tests in the future since we need to check a lot of different cases with this injection (this is why I created a specific assets folder for nb-injection)

CHANGELOG.md
edublancas: yeah the dev stays until we make a new release, once I run the release process 0.21.1dev becomes 0.21.1 (or 0.22 in case where making a major release)

idomic (approved):
lgtm

PR #986, opened 2022-08-21
Change IPython display to use IFrame

### LINKED ISSUE #842: use IFrame to display embedded D3 plot
when generating a pipeline plot with the D3 backend with `output='embed'`:

https://github.com/ploomber/ploomber/blob/77d7dbf6b143ee17f886f76b6f3163d0ad5a39c9/src/ploomber/dag/dag.py#L893

we use the requests-html library to load the HTML file since we need to execute the embedded JS for the plot to display:

https://github.com/ploomber/ploomber/blob/77d7dbf6b143ee17f886f76b6f3163d0ad5a39c9/src/ploomber/dag/plot.py#L66

and then we render the plot with a `IPython.display.HTML` object, this allows us to embed the plot in Jupyter. However, as pointed in https://github.com/ploomber/projects/issues/42 IPython has an `IFrame` object that correctly displays HTML and runs JS.

We should remove the requests-html code and substitute it for using `IFrame`

pelincetin:
## Describe your changes

I replaced the IPython display to use IFrame and edited the tests. 

Closes #842 

### DIFF
--- a/src/ploomber/dag/plot.py
+++ b/src/ploomber/dag/plot.py
@@ -4,7 +4,7 @@ from importlib.util import find_spec
 from pathlib import Path
 
 import jinja2
-from IPython.display import HTML
+from IPython.display import IFrame
 
 try:
     import importlib.resources as importlib_resources
@@ -71,27 +71,10 @@ def with_d3(graph, output, image_only=False):
         Path(output).write_text(rendered)
 
 
-@requires(['requests_html', 'nest_asyncio'],
-          name='embedded HTML with D3 backend',
-          pip_names=['requests-html', 'nest_asyncio'])
 def embedded_html(path):
-    import asyncio
 
     import nest_asyncio
+    
     nest_asyncio.apply()
-    return asyncio.get_event_loop().run_until_complete(
-        _embedded_html(path=path))
 
-
-async def _embedded_html(path):
-    # https://github.com/jupyter/nbclient/blob/1d629b2bed561fde521e6408e190a8159f117ddc/nbclient/util.py
-    # https://github.com/jupyter/nbclient/blob/main/requirements.txt
-    from requests_html import HTML as HTML_
-    from requests_html import AsyncHTMLSession
-
-    session = AsyncHTMLSession()
-    html = HTML_(html=Path(path).read_text(), session=session)
-    await html.arender()
-    # ensure we close the session, otherwise this will fail on windows
-    await session.close()
-    return HTML(data=html.html)
+    return IFrame(src=path, width=700, height=600)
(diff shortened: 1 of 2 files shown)


src/ploomber/dag/plot.py
```diff
@@ -71,27 +69,10 @@ def with_d3(graph, output, image_only=False):
...
-    from requests_html import AsyncHTMLSession
-
-    session = AsyncHTMLSession()
-    html = HTML_(html=Path(path).read_text(), session=session)
-    await html.arender()
-    # ensure we close the session, otherwise this will fail on windows
-    await session.close()
-    return HTML(data=html.html)
+
+    return IFrame(src=path, width=700, height=600)
```
pelincetin, line 78: Took these numbers from https://github.com/ploomber/projects/issues/42

tests/dag/test_dag.py
```diff
@@ -212,29 +211,12 @@ def test_plot_validates_html_extension_if_d3(dag, tmp_directory):
...
-
-    with pytest.raises(ImportError) as excinfo:
-        dag.plot(backend='d3')
-
-    expected = ("'requests-html' 'nest_asyncio' are "
-                "required to use 'embedded HTML with D3 backend'. "
-                "Install with: pip install 'requests-html' 'nest_asyncio'")
-    assert expected == str(excinfo.value)
+    # test path exists
+    assert len(output.src) != 0
```
edublancas, line 219: here we should test `assert Path(output.src).is_file()`. I ran the code, and it's breaking because we're removing the file when we bass the `"embed"` option, which is the default.

we have to modify this: https://github.com/ploomber/ploomber/blob/9c296ecfde0c8359631f8b09e6a8f4507e95863f/src/ploomber/dag/dag.py#L933 so it doesn't delete the file

edublancas (changes requested):

tests/dag/test_dag.py
yafimvo: Disabling the deletion doesn't solve the issue entirely since the temp file is being saved in `{temp-folder}/tmp` folder and when I use Jupyter notebook server it fails to access and embed the file with the following error: `Not allowed to load local resource: <URL>` 

Saving the file in the working dir solves this issue. 

Is this solution sounds OK? 
The main downside here is that every time we'll run the `plot` command (when the format of the file is `html`) it will generate a file.

edublancas: where did you test it? I think Python creates a temp file in a directory where it can read and write, but I might be wrong.

yafimvo: I've tested it on windows by running `jupyter notebook`. It creates a tmp file in default directory i.e : 'C:\Users\User\AppData\Local\Temp\tmp1p6cdvl7.html'.
The file exists but jupyer can't access it. If I move the file to the folder where I'm running the notebook it works.

image

image

edublancas: I see; ok, let's create it locally then.

edublancas: in the current working directory

PR #988, opened 2022-08-22
added suffix flag for automatically add number suffix to products

### LINKED ISSUE #973: missing detail in grid documentation
in the docs we have an example that looks like this:

```yaml
tasks:
  - source: random-forest.py
    name: random-forest-
    product: 'n_estimators=[[n_estimators]]/criterion=[[criterion]].html'
    grid:
        n_estimators: [5, 10, 20]
        criterion: [gini, entropy]

```

[link](https://docs.ploomber.io/en/latest/api/spec.html#tasks-grid)

the `[[placeholders]]`` are replaced at runtime by the parameter values, however, to ensure each file has a different name we also append a number at the end; but this isn't documented.

right below the snippet that I put above, we should add a little note saying that we number the filenames to prevent conflicting paths.

Wxl19980214:
## Describe your changes
User now can add grid_number_suffix to pipeline.yaml
With it on we automatically add number suffix to outputs to prevent same names;
With it off, we print error if output contain same names, user need to turn it on or have proper amount of placeholders.

Doc added.

Closes #973 

### DIFF
--- a/doc/api/spec.rst
+++ b/doc/api/spec.rst
@@ -954,6 +954,32 @@ for example, it will store the random forest with ``n_estimators=5``, and
 uses square brackets to differentiate them from regular placeholders when
 using an ``env.yaml`` file.
 
+In the latest version, we have introduced a speical flag called `grid_number_suffix`.
+The purpose of this flag is to prevent naming conflicts of our outputs. It is default 
+to be turned off. For example, in above example, since we have two placeholders for 
+two params in the grid, all products will have different names, thus it works properly.
+
+Consider this case: 
+
+.. code-block:: yaml
+    :class: text-editor
+    :name: grid-example-2-yaml
+
+    tasks:
+      - source: random-forest.py
+        name: random-forest-
+        product: 'n_estimators.html'
+        grid:
+            n_estimators: [5, 10, 20]
+            criterion: [gini, entropy]
+
+All products will have the same `n_estimators.html` name and that is a problem. Such pipeline.yaml
+will output error to indicate either we can add `grid_number_suffix: True`, or we
+can give proper placeholders like the other case. If we turned on the flag, the output files will be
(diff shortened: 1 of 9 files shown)


idomic:
Add a changelog entry as well?

doc/api/spec.rst
```diff
@@ -954,6 +954,32 @@ for example, it will store the random forest with ``n_estimators=5``, and
 uses square brackets to differentiate them from regular placeholders when
 using an ``env.yaml`` file.
 
+In the latest version, we have introduced a special flag called `grid_number_suffix`.
```
edublancas, line 957: please add a specific version here (this would be 0.21.1) and add an entry to the changelog under this section, see this: https://github.com/ploomber/ploomber/blob/master/CONTRIBUTING.md#documenting-changes-and-new-features

Wxl19980214: Added version and updated docstring.

idomic:
@Wxl19980214 feel free to fix the errors and tag when ready +1

Wxl19980214:
i think its ready

idomic:
Ci is failing

Wxl19980214:
Oh yeah, i thought it was an github error. Fixed it, it should pass now.

Wxl19980214:
```
./src/ploomber/telemetry.py:6:80: E501 line too long (84 > 79 characters)
```
Wait this is weird, I run flake8 locally and it did not return this error? Also I think that line only has 64 characters?

edublancas:
>  Wait this is weird, I run flake8 locally and it did not return this error? Also I think that line only has 64 characters?

did you rebase? I introduced this error and I indeed saw that the line was >80, it's fixed on master

edublancas:
closing in favor of https://github.com/ploomber/ploomber/pull/1020

PR #989, opened 2022-08-22
Task default parameter

### LINKED ISSUE #648: Setting task class default init parameters

Wxl19980214:
draft pr, will fail.

closes #648

### DIFF
--- a/src/ploomber/spec/dagspec.py
+++ b/src/ploomber/spec/dagspec.py
@@ -444,6 +444,7 @@ class DAGSpec(MutableMapping):
                 'config',
                 'clients',
                 'tasks',
+                'task_defaults',
                 'serializer',
                 'unserializer',
                 'executor',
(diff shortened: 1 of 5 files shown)


idomic:
@Wxl19980214 are you still on this?

idomic:
Closing due to inactivity, feel free to open.

Issue #982, opened 2022-08-17
Windows tests failing `tests/test_requests.py::test_timeout_error[get]`


grnnja:
Sometimes the windows tests fail with this test not passing:
`FAILED tests/test_requests.py::test_timeout_error[get]`

This can be seen here:
https://github.com/ploomber/ploomber/runs/7885886283?check_suite_focus=true#step:8:853

PR #991, opened 2022-08-23
Script for easy installation

### LINKED ISSUE #939: script for easy installation
Many users get stuck while trying to install ploomber. When checking it out for the first time, many go to the terminal and run `pip install ploomber`.  However, if they have an existing environment with lots of packages, they're running on windows, among other factors. Chances are, the pip command will fail.

A script that takes care of this can significantly simplify the process for users. We can do this with a Python script, this will make it easier to embed logic, and it's more portable (we want this to work well on windows). Initially, we can aim to get it working on Python 3.7 and over. Note that given the nature of this task, the script should only use the standard library and not any external packages.

## basic logic

* check Python version >= 3.7
* if conda is available, use it to create the virtual environment
* if conda is not available, use [venv](https://docs.python.org/3/library/venv.html) 
* install ploomber
* print follow-up steps (include command on for activating the virtual env)

Much of this logic is already here, so we can use it as reference: https://github.com/ploomber/ploomber/blob/master/src/ploomber/cli/install.py

## some other ideas

* if docker is available, this can simplify things a lot, but maybe for v2
* to know if this is successful, we should implement some anonymous telemetry (not sure what's the easiest option here). perhaps we can create a netlify app, hit a URL with different parameters and then look at the analytics. we want to record: python version, if docker is available, and if the script was successful or not
* we could go beyond installing ploomber and start jupyterlab, check that the jupyterlab extension works, and maybe even starting jupyterlab with an interactive tutorial (see this https://github.com/ploomber/ploomber/issues/936)

e1ha:
## Describe your changes
Added a shell script that would download python if it doesn't exist and a python script that checks python version and installs ploomber in a virtual environment either conda or python depending on if they have conda.
Closes #939 


e1ha:
I wasn't sure what the best way to test this would be. I tested that it would run on my laptop via the command line by doing sh ./src/ploomber/cli/installPloomber.sh while in the ploomber directory and from my tests, it worked.

edublancas:
this a good start. a couple of notes:

1. installing homebrew might take a while; I think it's better to install [miniconda](https://docs.conda.io/en/latest/miniconda.html); this will allow us to install a Python env as well. the conda install has some user prompts, but I remember there's a way to pass them as command line flags, so it doesn't ask for them
2. after installing ploomber, we need to handhold them even more. I think we can do this: install JupyterLab, copy one example (by running `ploomber examples`, and start jupyterlab. ultimately, we want a ready-to-use environment so they start working locally.
3. something that can happen is that users are already running JupyterLab (and executed this from a terminal inside JupyterLab), so I think we could check if the command is executed inside jupyter (and skip launching it). perhaps with some environment variable?


thoughts @idomic @e1ha ?

e1ha:
1. For 1, I'm looking into how to install conda via the command line. From my research online, it seems to be a mixture of wget and bash for linux and macOS but that would be dependent on the architecture, which would vary across users. Would you know of an easier way?

2. I have a question about 2, specifically this part:
>  I think we can do this: install JupyterLab, copy one example (by running ploomber examples, and start jupyterlab.
Why do we need to copy an example? If we're testing JupyterLab wouldn't starting JupyterLab be sufficient?

idomic:
I agree here, already wrote a DM about brew.
We have to get an example with the jupyter so you have it ready once it opens (one less action the user has to do).

e1ha:
I'm still a bit confused with the example part. Would I run `ploomber examples` and then run `jupyter lab` after that?

idomic:
Exactly. Getting the intro-to-ploomber example and then running Jupyter lab
from its directory.

src/ploomber/cli/install_ploomber.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
+
+which pythona > /dev/null 2>&1
+PYTHON_EXIT_STATUS=$?
+if [ "$PYTHON_EXIT_STATUS" -ne "0" ] 
+then
+    echo 'Python not found. Installing Python...'
+
+    which condaa > /dev/null 2>&1
```
idomic, line 9: typo?

idomic, line 9: typo?

idomic (commented):
Tag us when ready for review?

src/ploomber/cli/install_ploomber.sh
e1ha: Yeah that's a typo. I was testing the which command. I'll tag you when it's done since the CI/CD pipeline is still failing.

e1ha:
@idomic It's done now. A few comments: 
- I couldn't test that either scripts would work on windows or Linux since I don't have windows or Linux but it works on my mac. If there is a way to test it on Linux or Windows let me know please.
- The way I tested it is that I ran the script and it worked. It opens Jupyter Lab at the end and when I do which jupyter-lab and which ploomber, the jupyterlab and ploomber packages are in the environment. 
- I ran the shell script in a virtual environment. I removed conda and it was able to be installed. I also did which python and it was installed in the virtual environment.
- The shell script requires the user to have the ploomber repo. How should I make it so that the user wouldn't need to have the ploomber repo in order to run the shell script?

e1ha:
What is the shell that the CI/CD pipeline is running on?

e1ha:
CI/CD pipeline is failing on this error 
```
@pytest.mark.parametrize('method', methods)
    def test_timeout_error(method):
        with pytest.raises(exceptions.NetworkException) as excinfo:
>           method('https://ploomber.io/', timeout=0.0000001)
E           Failed: DID NOT RAISE <class 'ploomber.exceptions.NetworkException'>
```
I don't think that has anything to do with my script but I could be wrong.

e1ha:
Should I write tests for this script?

e1ha:
Also for the error that is failing, Prem mentions that sometimes they don't pass on this issue #982 .

e1ha:
@idomic all CI/CD pipeline issues have been fixed, so this PR is ready for review.

PR #1000, opened 2022-08-25
add tabs for cli/python api

### LINKED ISSUE #938: tabs in the documentation to show CLI and Python API
Currently, all our documentation shows how to use ploomber via the CLI. However, many users might prefer to use a Jupyter notebook to interact with Ploomber instead of a terminal. In such cases, it's better for them to use the Python API since it provides more excellent output. Users aren't finding this feature, so it'd be best to include the CLI and Python examples in each documentation section. Something like this:

For example. When we show that users can build with `ploomber build`. The equivalent in Python is:

```python
# prepare
from ploomber.spec import DAGSpec
dag = DAGSpec.find().to_dag()

# build
dag.build()
```

Note that the `prepare` part is necessary for running any command. So the "Python API tab" should contain a link with a short tutorial explaining how to get the `dag` variable by executing the "prepare" part, and the tab only contain the actual command (e.g. `dag.build()`)

### ipynb examples

some of the examples in our documentation are ipynb files pulled from the [projects repository](https://github.com/ploomber/projects). The challenge is that the libraries for adding tabs won't work there. So we need to figure out a way to have executable examples where we can embed this logic. My recommendation is to check the resources of executable books. Based on some quick research, it looks like we can write notebooks in MyST format and still be able to use sphinx plugins. My guess is that we do not need to completely change our documentation and only adopt the [MyST parser](https://github.com/executablebooks/MyST-Parser).

### Python API statefulness

When using the CLI, each command executes in a new process, so there isn't any state saved between calls. However, when using the Python API, things might behave unexpectedly.

For example, if a user runs `dag.build()`, modifies a task's source and runs `dag.build()` again - the source code change *wont be detected*, this is because once we load the DAG, we don't check status changes. An (undocumented) hot reload feature works reasonably well, but we should ensure it works correctly before showing the Python API snippets.

Relevant links to check activate the hot reload feature:

https://github.com/ploomber/ploomber/blob/2c5417abb606a83d441737acf6e4ac3877364ac5/src/ploomber/dag/dag.py#L233
https://github.com/ploomber/ploomber/blob/master/src/ploomber/dag/dagconfiguration.py

grnnja:
## Describe your changes

here is an example:

myst
````md
~~~{ploomber-example}
  ```{example-python}
      ```sh magic_args="--no-raise-error"
      ploomber build --force
      ```
   ```
   ```{example-cli}
      ```sh magic_args="--no-raise-error"
      ploomber build --force
     ```
   ``` 
~~~
````

equivalent to rst:
```rst
.. ploomber-example::

   .. example-python::

       .. code-block:: Bash
          :class: text-editor
          :name: task-bash

          ploomber build --force
   .. example-cli::

       .. code-block:: bash
          :class: text-editor
          :name: task-bash

          ploomber build --force

```

rendered:
image
image

As for Python API statefulness I think we just have to keep that in mind when writing the docs.

Closes #938

### DIFF
--- a/doc/_ext/ploomber_example.py
+++ b/doc/_ext/ploomber_example.py
@@ -1,0 +1,92 @@
+from docutils import nodes
+from docutils.statemachine import ViewList
+from sphinx_tabs.tabs import TabsDirective, GroupTabDirective
+
+
+class PloomberExampleCliDirective(GroupTabDirective):
+
+    has_content = True
+
+    def run(self):
+
+        self.assert_has_content()
+
+        new_content = ViewList([
+            'CLI',
+            '',
+        ])
+
+        new_content.append(self.content)
+
+        self.content = new_content
+
+        node = super().run()
+
+        return node
+
+
+class PloomberExamplePythonDirective(GroupTabDirective):
+
+    has_content = True
+
+    def run(self):
+
+        self.assert_has_content()
+
+        # ploomber example config
+        # found with python -m sphinx.ext.intersphinx _build/html/objects.inv
+        ploomber_example_prepare_info_path = '../user-guide'\
+            '/using-python-api.html'
+
+        ref_node = nodes.reference(
+            '',
+            '',
+            internal=True,
+            # refname=app.config["ploomber_example_prepare_info_path"])
+            refuri=ploomber_example_prepare_info_path)
+
+        inner_node = nodes.emphasis("here.", "here.")
+
+        ref_node.append(inner_node)
+
+        note_node = nodes.note()
+
+        warning_text_node = nodes.paragraph(
+            text='Unlike the command line interface, the Ploomber Python '
+            'API requires preparation before each command. Find more '
+            'information ')
+
+        warning_text_node += ref_node
+
+        note_node.append(warning_text_node)
+
+        new_content = ViewList([
+            'Python',
+            '',
+        ])
+
+        new_content.append(self.content)
+
+        self.content = new_content
+
+        tab_node = super().run()[0]
+
+        # prepend
+        tab_node.insert(0, note_node)
+
+        return [tab_node]
+
+
+def setup(app):
+    app.add_config_value('ploomber_example_prepare_info_path',
+                         'unconfigured prepare guide path', 'html')
+
+    app.add_directive("ploomber-example", TabsDirective)
+    app.add_directive("example-python", PloomberExamplePythonDirective)
+    app.add_directive("example-cli", PloomberExampleCliDirective)
+
+    return {
+        'version': '0.1',
+        'parallel_read_safe': True,
+        'parallel_write_safe': True,
+    }
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -15,6 +15,7 @@ import sys
 from ploomber import __version__
 
 sys.path.insert(0, os.path.abspath('.'))
+sys.path.append(os.path.abspath('./_ext'))
 import hooks  # noqa
 
 # -- Project information -----------------------------------------------------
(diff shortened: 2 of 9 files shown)


idomic:
@e1ha please submit your code review.

doc/_ext/ploomber_example.py
```diff
@@ -0,0 +1,92 @@
...
+            refuri=ploomber_example_prepare_info_path)
+
+        inner_node = nodes.emphasis("here", "here")
+
+        ref_node.append(inner_node)
+
+        note_node = nodes.note()
+
+        warning_text_node = nodes.paragraph(
+            text='Unlike the command line interface, the Ploomber Python '
```
e1ha, line 55: This might not matter but the two sentences are missing periods.

doc/user-guide/logging.md
```diff
@@ -0,0 +1,154 @@
+<!-- start header -->
```
idomic, line 1: So why are we changing all of these guides into markdown?

idomic (changes requested):
I also didn't see where was the myst applied?
Might be worth applying to a single doc and see how it works.

doc/user-guide/logging.md
grnnja: Just changed it to ignore .md files in subdirectories in the doc directory. So we aren't committing all the guides now

doc/_ext/ploomber_example.py
grnnja: fixed!

doc/conf.py
```diff
@@ -31,12 +32,9 @@
...
 extensions = [
-    'sphinx.ext.autodoc',
-    'sphinx.ext.napoleon',
-    'sphinx.ext.autosummary',
-    'sphinx.ext.autosectionlabel',
-    'nbsphinx',
-    'sphinx_toolbox.collapse',
+    'sphinx.ext.autodoc', 'sphinx.ext.napoleon', 'sphinx.ext.autosummary',
+    'sphinx.ext.autosectionlabel', 'nbsphinx', 'sphinx_toolbox.collapse',
+    'sphinx_tabs.tabs', 'myst_parser', 'ploomber_example'
```
grnnja, line 37: > I also didn't see where was the myst applied?

applied here

grnnja:
>  Might be worth applying to a single doc and see how it works.

Not sure what you mean by this. If you want to try out the changes if you copy the rst from the first comment into one of the doc pages and rebuild you can see the changes

e1ha:
It might be worth putting the example in some form of documentation so other users know how to use it since there are no examples currently.

grnnja:
> It might be worth putting the example in some form of documentation so other users know how to use it since there are no examples currently.

Good point. I added information on using the directive in CONTRIBUTING.md

yafimvo (commented):
I'm not sure if I used it correctly but after cloning and running `invoke setup` (also `invoke setup --doc`) I was unable to run `invoke docs` because these libraries were missing: `nbsphinx`, `sphinx-toolbox`, and `myst-parser`.

After installing these dependencies manually and running docs - the tabs logic works and the code looks good to me.

It is also worth mentioning that I couldn't run `invoke setup` on Windows (using shell/cmd) in ./projects-ploomber/ inorder to run `conda activate projects` (image with err attached). 

* It works with bash

shell and cmd error:
image

grnnja:
Thanks for finding this. I added `myst-parser` to the environment.yml.

 It's strange that nbsphinx and sphinx-toolbox weren't installed with `invoke setup --docs` though since they are already ./doc/environment.yml. I'm also not sure what to do with the windows error

idomic:
@yafimvo please open an issue on this eval if doesn't exist, I think it's just a command that's platform specific to linux/bash. It should be working for windows as well, but not related to this PR.

idomic:
@yafimvo please review now that the dependency is in there

edublancas:
closing this in favor of https://github.com/ploomber/ploomber/pull/1028

PR #1003, opened 2022-08-26
Cli hiding traceback

### LINKED ISSUE #794: CLI hiding the error traceback
When a user operation triggers an error, we only print the error message if the exception is a `BaseException` (or a subclass):

https://github.com/ploomber/ploomber/blob/0a5fed2f9ec539c744664b1b7b3a7f744ec6c482/src/ploomber/cli/io.py#L99

However, when it's not a subclass, we want to show all the traceback, as it contains important details to debug. But the current implementation only prints the error message:

https://github.com/ploomber/ploomber/blob/0a5fed2f9ec539c744664b1b7b3a7f744ec6c482/src/ploomber/cli/io.py#L102

This makes debugging harder, see this https://github.com/ploomber/ploomber/issues/789#issuecomment-1130066043

shizuchanw:
## Describe your changes
added code to show full traceback when raising a non-subclass exception.
Closes #794 

### DIFF
--- a/src/ploomber/cli/io.py
+++ b/src/ploomber/cli/io.py
@@ -92,8 +92,6 @@ def cli_endpoint(fn):
     return wrapper
 
 
-# FIXME: capture only certain types of exceptions. If it's something we didn't
-# raise, we'd like to see the full traceback
 def command_endpoint(fn):
     """
     Decorator for command line endpoints that only parse dags or tasks but do
@@ -108,8 +106,14 @@ def command_endpoint(fn):
         except BaseException as e:
             click.secho(e.get_message(), file=sys.stderr, fg='red')
             sys.exit(1)
+        # show the full traceback if it's not a subclass Exception
         except Exception as e:
-            print(f'Error: {e}', file=sys.stderr)
+            error = _format.exception(e)  # get the traceback
+            if error:
+                tw = TerminalWriter(file=sys.stderr)  # write to terminal all the traceback
+                tw._write_source(error.splitlines())
+            else:
+                print(f'Error: {e}', file=sys.stderr)
             sys.exit(1)
 
     return wrapper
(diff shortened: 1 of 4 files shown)


edublancas:
we're missing a test, please create a `test_io.py` here: https://github.com/ploomber/ploomber/tree/master/tests/cli

then, let's run a a few test cases, by using `command_endpoint`: https://github.com/ploomber/ploomber/blob/0a5fed2f9ec539c744664b1b7b3a7f744ec6c482/src/ploomber/cli/io.py#L89

something like this:

```
def test_command_endpoint_shows_full_traceback():
    @command_endpoint
    def function():
        raise ValueError
    
    # check that the full traceback is displayed
```

This will raise a SystemExit, so you need to catch it: https://docs.python.org/3/library/exceptions.html#SystemExit

you also need to capture the standard error.

see this:

https://docs.pytest.org/en/7.1.x/how-to/capture-stdout-stderr.html
and this
https://docs.pytest.org/en/7.1.x/how-to/assert.html

idomic:
@shizuchanw please make sure the PR is passing the CI, I reran it.

shizuchanw:
> CI

Is CI referring to the unit tests here? 

I checked which tests weren't passing: `tests/test_requests.py::test_timeout_error[get]` and `tests\cli\test_examples.py::test_error_if_git_clone_fails`.
The first test is related to #982 , and the second test seems like a network error. Should I open an issue for the second test?
@idomic

shizuchanw:
update: the test `“test_error_if_git_clone_fails”` in tests/cli/test_examples.py  was failing because It asserted that the error message should be equal to the message printed, while I added full traceback to be printed in addition to that message. Therefore I changed test's equal assertion to "assert (error message) in stderr"

edublancas (approved):

edublancas:
thanks for taking care of this! 🎉

PR #1009, opened 2022-08-31
Dev/elvis

### LINKED ISSUE #789: Make ploomber build write parameters on source code as well
I've just installed `ploomber` to use in a simple personal project, but I've struggled with how the pipeline configuration in `pipeline.yaml` forces the user to leave some of the parameters in the YAML file and others in the script itself if the user wants the scripts to run manually, without using `ploomber`. I believe if `ploomber build` or some other command in the CLI also injected the parameters into the source code, this would not be an issue.

The use case I've imagined is this: a user wants to centralise all their parameters on the `pipeline.yaml` file so that they don't need to look for parameters in each of their source files, yet they also want the option to run their source code manually, without using `ploomber`. Currently, it seems to me this is not possible, and a concrete example from the templates highlights this: it is not possible to run the [`clean.py`](https://github.com/ploomber/projects/blob/de47cf98cc6b4cc14117924b623ded989ab870d7/templates/exploratory-analysis/scripts/clean.py) script from the `exploratory-analysis` template by itself, since the parameters tag states that `product = None` and `ploomber` does not inject the parameters back into the source code when running `ploomber build` or any other command. Ideally, you'd have another command in the CLI such as `ploomber sync` that would take the parameters from `pipeline.yaml` and write them back to the source code.

The same problem is also present when using functions as source, though I can't think of any way for allowing scripts to be executed by themselves that would make sense -- just running e.g `python script.py` feels weird. Also I'm not sure about notebooks in the `.ipynb` format, because the videos I've seen seem to imply the parameter cells are added to them, though only when editing on Jupyter?

tianyiyu-mcit:
## Describe your changes
added method to track traceback of exceptions
Closes #789

- [x ] I have performed a self-review of my code


idomic:
Try to open a clean pr and cherry pick your changes, it looks like you merged master into your branch so it's hard to see what changed.

PR #1013, opened 2022-09-01
Dev/elvis

### LINKED ISSUE #794: CLI hiding the error traceback

tianyiyu-mcit:
## Describe your changes
Added tracking traceback for exception if the exception is not a subclass of BaseException
Closes #794 

### DIFF
--- a/src/ploomber/cli/io.py
+++ b/src/ploomber/cli/io.py
@@ -109,7 +109,7 @@ def command_endpoint(fn):
             click.secho(e.get_message(), file=sys.stderr, fg='red')
             sys.exit(1)
         except Exception as e:
-            print(f'Error: {e}', file=sys.stderr)
+            print(traceback.format_exc())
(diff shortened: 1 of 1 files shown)


edublancas:
can you add a unit test?

PR #1014, opened 2022-09-02
Created new button for side bar

### LINKED ISSUE #366: HTML/CSS - Fix right bar in docs
When opening in a large monitor, the docs display a right bar to move between sections ([example](https://ploomber.readthedocs.io/en/latest/user-guide/deployment.html)):

Screen Shot 2021-10-15 at 12 24 27

However, on tablets and phones, we remove that bar. It'd be better to add a button to the top bar to toggle it, just like we do with the left bar (see right button):

Screen Shot 2021-10-15 at 12 25 33

source code: https://github.com/ploomber/ploomber/blob/76b2abf78092e6696d9e25e47cefdc1e1589529f/doc/_templates/layout.html#L41

check doc/contributing.md for instructions

e1ha:
## Describe your changes
Create a new button that will toggle the side bar. 

Closes #366 

### DIFF
--- a/doc/_static/css/custom-theme.css
+++ b/doc/_static/css/custom-theme.css
@@ -183,6 +183,15 @@ pre> :not(:first-child).ipynb-lexed * {
         margin-left: 50px;
         margin-bottom: 10px;
     }
+    #third-column {
+        position: fixed !important;
+        top: 3.5rem;
+        left: 0;
+        width: 100%;
+        height: 100%;
+        z-index: 100;
+        background-color: #fff;
+    }
 }
 
 .sphinxsidebarwrapper ul {
--- a/doc/_templates/layout.html
+++ b/doc/_templates/layout.html
@@ -111,7 +111,7 @@ titles_only=false) %}
                 <div class="clearer"></div>
             </div>
         </div>
-        <div class="d-none d-xl-block col-xl-2 third-column">
+        <div class="d-none d-xl-block col-xl-2 third-column" id="third-column">
             {{ sidebar2() }}
         </div>
     </div>
--- a/doc/_templates/macros.html
+++ b/doc/_templates/macros.html
@@ -8,7 +8,14 @@
         </svg>
     </button>
 
-    <a class="navbar-brand" href="{{ pathto('index') }}">Ploomber</a>
+    <a class="navbar-brand ml-auto mr-auto" href="{{ pathto('index') }}">Ploomber</a>
+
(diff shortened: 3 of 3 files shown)


e1ha:
This PR is done. For the error that is failing on the CI/CD pipeline, Prem mentions that sometimes they don't pass on this issue #982 . What should I do to fix it?

e1ha:
This PR is ready for review.

grnnja (commented):
It's pretty confusing to have three unlabeled buttons at the top like this: 
image

@e1ha @edublancas 
What do you think about showing the outline before the title of the page with a title of "Outline v" then the outline, and if the user presses on "Outline" or the down arrow it collapses to just "Outline >" Somewhat similar to hackernews:

image
image

Also right now if you press on one of the links in the menu it doesn't give feedback that anything has happened. I would expect it to close and jump to the link

yafimvo (commented):
Checked it locally. Works and the code looks right.

It's possible to improve code reuse and readability by making a few minor changes

1. Naming: 
#third-column can be something like #top-sidebar

2. Re-use toggle functionality
instead of this method

```
function toggleThirdColumn() {
    let thirdColumn = document.getElementById("third-column");
        if (thirdColumn.classList.contains("d-none")) {
            thirdColumn.classList.remove("d-none");
        } else {
            thirdColumn.classList.add("d-none");
        }
} 
```
we can split the logic and create a toggle function that can be used as a utility later
```
function toggleTopSidebar() {
    let sidebar = document.getElementById("third-column");
    toggleElementClass(sidebar, "d-none")
}

function toggleElementClass(element, className) {
    if (element.classList.contains(className)) {
        element.classList.remove(className);
     } else {
        element.classList.add(className);
     }
}
```

yafimvo:
@idomic @grnnja 

When the sidebar is open it covers the entire screen which may affect the user experience when trying to navigate within a doc.

i.e: Clicked index 
image


Instead making `height: 100%` we can do `height: min-content` so while the menu is open, the doc's content will be visible
image

yafimvo:
Agree. I've separated this point from the code review and added it as a general suggestion.

grnnja:
Didn't really know how to design this to make it look ok, but this is what I came up with. I'll make another pr addressing the buttons in the top bar.
image
image
image

idomic:
@yafimvo please review this PR

idomic:
closed by #1031

PR #1019, opened 2022-09-12
Prioritized cell injection

### LINKED ISSUE #902: Injecting parameters when the same notebook appears more than once

yafimvo:
## Describe your changes

Use `setup.cfg` to set which parameters to inject.

setup.cfg
```
[ploomber]
entry-point = pipeline.yaml
inject-priority = task-a
```

pipeline.yaml
```
tasks:
    - source: template_a.ipynb
      name: task-a
      product: output/template-a.ipynb
      params:
        some_param: param-a

    - source: template_a.ipynb
      name: task-a-suffix
      product: output/template-a-suffix.ipynb
      params:
        some_param: param-a-suffix

    - source: template_b.ipynb
      name: task-b
      product: output/template_b.ipynb
      params:
        some_param: param-b

    - source: template_b.ipynb
      name: task-b-suffix
      product: output/template-b-suffix.ipynb
      params:
        some_param: param-b-suffix
```

Use wildcards to inject multiple parameters (*)

config.cfg
```
[ploomber]
entry-point = pipeline.yaml
inject-priority = *-suffix
```

This PR is based on [PR 981](https://github.com/ploomber/ploomber/pull/981) and [PR 985](https://github.com/ploomber/ploomber/pull/985)

Closes #902 


idomic:
@yafimvo here I see there are some unit tests that are failing.

edublancas:
I still need to test this, but I don't see any updates to the documentation. has this been documented?

yafimvo:
Since it was part of the `setup.cfg` file, I updated the changelog. Where should I add this documentation?

edublancas:
please add the docs at the bottom of this section: https://docs.ploomber.io/en/latest/get-started/basic-concepts.html#the-cell-injection-process

edublancas (approved):

PR #1020, opened 2022-09-12
973 missing detail in grid documentation

### LINKED ISSUE #973: missing detail in grid documentation

shizuchanw:
## Describe your changes
based on [PR: 988](https://github.com/ploomber/ploomber/pull/988), fixed flake8 issue and checked documentation.

Closes #973 

### DIFF
--- a/src/ploomber/spec/taskspec.py
+++ b/src/ploomber/spec/taskspec.py
@@ -299,6 +299,7 @@ class TaskSpec(MutableMapping):
 
         if 'grid' in data:
             data_source_ = data["source"]
+            grid_number_suffix = data.pop('grid_number_suffix', False)
             data_source = str(data_source_ if not hasattr(
                 data_source_, '__name__') else data_source_.__name__)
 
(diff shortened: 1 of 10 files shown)


edublancas:
so this is a replacement of https://github.com/ploomber/ploomber/pull/988, right?

shizuchanw:
> so this is a replacement of #988, right?

Yes, I wasn’t able to directly commit to #988 so I created a new branch and copied the changes.

src/ploomber/spec/taskspec.py
```diff
@@ -299,6 +299,7 @@ def to_task(self, dag):
 
         if 'grid' in data:
             data_source_ = data["source"]
+            grid_number_suffix = data.pop('grid_number_suffix', False)
```
edublancas, line 302: the default should be True, so we dont break existing pipelines. please change the default, update the tests (if anything breaks) and update the documentation

edublancas (changes requested):

PR #1028, opened 2022-10-06
Cli and python tabs in docs

### LINKED ISSUE #938: tabs in the documentation to show CLI and Python API

yafimvo:
## Describe your changes
Fixed sphinx dependencies issues

[Original PR](https://github.com/ploomber/ploomber/pull/1000#issue-1351165622)

Closes #938  

### DIFF
--- a/doc/hooks.py
+++ b/doc/hooks.py
@@ -36,9 +36,9 @@ def config_init(app, config):
 
     # move README.ipynb files to their corresponding location in the docs
     for path, target_dir in directories.items():
-        src = Path(projects, path, 'README.ipynb')
+        src = Path(projects, path, 'README.md')
         name = Path(path).name
-        dst = Path(target_dir, f'{name}.ipynb')
+        dst = Path(target_dir, f'{name}.md')
         print(f'Copying {src} to {dst}')
         shutil.copy(src, dst)
 
--- a/doc/user-guide/using-python-api.rst
+++ b/doc/user-guide/using-python-api.rst
@@ -1,0 +1,37 @@
+.. _Using-Python-In-Docs
+Using the Python API in the docs
+===============================
+
+Unlike the command line interface, the Ploomber Python
+API requires this preparation before each command:
+
+.. code-block:: python
+    :class: text-editor
+    :name: task-py
+
+    from ploomber.spec import DAGSpec
+    dag = DAGSpec.find().to_dag()
+
+
+For example, in building:
+
+.. code-block:: python
+    :class: text-editor
+    :name: task-py
+
+    # prepare
+    from ploomber.spec import DAGSpec
+    dag = DAGSpec.find().to_dag()
+
+    # build
+    dag.build()
+
+is equivalent to:
+
+.. code-block:: console
+
+    ploomber build
+
+For brevity the documentation does not include the prepare command in every code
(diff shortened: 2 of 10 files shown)


doc/hooks.py
```diff
@@ -36,9 +36,9 @@ def config_init(app, config):
 
     # move README.ipynb files to their corresponding location in the docs
     for path, target_dir in directories.items():
-        src = Path(projects, path, 'README.ipynb')
+        src = Path(projects, path, 'README.md')
         name = Path(path).name
-        dst = Path(target_dir, f'{name}.ipynb')
+        dst = Path(target_dir, f'{name}.md')
```
edublancas, line 41: what changed here? I remember we are explicitly using the ipynb files because they already contain the output cells, wouldn't this break stuff?

doc/user-guide/using-python-api.rst
```diff
@@ -0,0 +1,37 @@
+.. _Using-Python-In-Docs
+Using the Python API in the docs
+===============================
```
edublancas, line 3: this looks like a document that we want to embed whenever we display the tabs right? I'd say let's move it to the FAQ instead of being a complete section.

Also, I think it'd be good to automatically embed this at the bottom of the Python API tab so we don't have to put this there every time

edublancas (changes requested):

PR #1033, opened 2022-10-12
Note on naming of product dict keys, typos

### LINKED ISSUE #888: clarify in the docs that keys in the product field can be anything
I got a user who asked me if their `pipeline.yaml` looked well and I noticed they had something like this:

```
tasks:
   # ... a bunch of tasks
   - source: fit.py
      product:
        nb: path/to/report.html
        data: path/to/data.csv
        data1: path/to/data.csv
        data2: path/to/data.csv
```

The user was confused and thought that the naming should be name, name1, name2. So we should add a little note in the basic concepts guide to explicitly say that keys can be anything

https://docs.ploomber.io/en/latest/get-started/basic-concepts.html

mhered:
## Describe your changes
* Add a note to clarify that the keys of product dictionaries can be given any name, in files  doc/get-started/basic-concepts.rst and doc/api/spec.rst

* Fixing minor typos in both files.
Closes #888 

- [X] I have added thorough tests (when necessary). --> not needed

### DIFF
--- a/doc/api/spec.rst
+++ b/doc/api/spec.rst
@@ -597,7 +597,9 @@ If the task generates multiple products, pass a dictionary:
             data: output/data.csv
 
 
-The mechanism to make ``product`` available when exeuting your task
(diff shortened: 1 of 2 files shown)


edublancas:
great work @mhered; thanks a lot!

PR #1038, opened 2022-10-21
Added media query to avoid horizontal scroll

### LINKED ISSUE #1022: Mobile Docs Page X Overflow
On docs pages with long words in the title we get overflow which make the page have horizontal scrolling.

galaxy s10:
image

I think if remove some padding and maybe decreased the font size on mobile it would fix this.

Nekros1712:
Reduced font size for small width devices
Closes #1022


edublancas:
I was unable to reproduce the reported issue #1022

PR #1039, opened 2022-10-21
Handling console error when right bar does not appear

### LINKED ISSUE #767: fix js when right bar does not appear
there is some JS code that highlights the currently active section in the right sidebar. See this: https://docs.ploomber.io/en/latest/user-guide/cli.html

However, not all sections in the docs show the right sidebar, in such case, the console keeps throwing an error:

image

Nekros1712:
Added a condition to check if there is at least a section to show then only popping from current list

Closes #767


edublancas (approved):

edublancas:
thanks for your contribution!

PR #1042, opened 2022-11-03
Add environment vars expansion and pre defined git hashes in the yaml…

feribg:
… file

Closes https://github.com/ploomber/soopervisor/issues/116

- [ x ] I have performed a self-review of my code
- [ x ] I have added thorough tests (when necessary).

### DIFF
--- a/tests/env/test_env.py
+++ b/tests/env/test_env.py
@@ -63,6 +63,21 @@ def test_env_repr_and_str_when_loaded_from_file(tmp_directory, cleanup_env,
     assert str(env) == str(d)
 
 
+def test_env_load_env_vars(tmp_directory, cleanup_env,
+                           monkeypatch):
+
+    path_env = Path('env.yaml')
+    d = {
+        'env_user': '{{env.USER}}',
+        'env_path': '{{env.PATH}}'
+    }
+
+    path_env.write_text(yaml.dump(d))
+    env = Env()
+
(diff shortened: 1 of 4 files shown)


tests/env/test_env.py
```diff
@@ -63,6 +63,21 @@ def test_env_repr_and_str_when_loaded_from_file(tmp_directory, cleanup_env,
     assert str(env) == str(d)
 
 
+def test_env_load_env_vars(tmp_directory, cleanup_env,
```
edublancas, line 66: please set some env vars using monkeypatch: https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.MonkeyPatch.setenv

and assert that they contain the appropriate values

tests/env/test_sample_project_expanders.py
```diff
@@ -42,6 +42,15 @@ def test_error_on_unknown_placeholder():
     assert expected == str(excinfo.value)
 
 
+def test_expand_env_vars():
+    expander = EnvironmentExpander({})
```
edublancas, line 46: same as the other test: add monkeypatch, set some env vars and check the values

edublancas (changes requested):
please add docs here: https://github.com/ploomber/ploomber/blob/master/doc/api/spec.rst#default-placeholders

edublancas (approved):

PR #1051, opened 2022-12-19
Fix #1048 - nb inject fails when using ploomber scaffolding for a package

### LINKED ISSUE #1048: nb inject fails when using ploomber scaffolding for a package
When using the [ploomber scaffolding for a package](https://docs.ploomber.io/en/latest/user-guide/scaffold.html#packaging-projects) my attempts to run ploomber nb --inject result in the following error:

```
  File "path-to-venv/venv/lib/python3.9/site-packages/ploomber/cli/io.py", line 103, in wrapper
    fn(**kwargs)
  File "path-to-venv/venv/lib/python3.9/site-packages/ploomber_core/telemetry/telemetry.py", line 545, in wrapper
    raise e
  File "path-to-venv/venv/lib/python3.9/site-packages/ploomber_core/telemetry/telemetry.py", line 533, in wrapper
    result = func(*args, **kwargs)
  File "path-venv/venv/lib/python3.9/site-packages/ploomber/cli/nb.py", line 472, in main
    params_to_inject, warning_message, exceptions = _get_params_to_inject(
  File "path-to-venv/venv/lib/python3.9/site-packages/ploomber/cli/nb.py", line 132, in _get_params_to_inject
    unprioritized_tasks_to_inject, warning_message = _get_tasks_to_inject(
  File "path-to-venv/venv/lib/python3.9/site-packages/ploomber/cli/nb.py", line 84, in _get_tasks_to_inject
    full_template_name = Path(task.source._path).name
AttributeError: 'PythonCallableSource' object has no attribute '_path'
```

Reprex:
```
> ploomber scaffold --package
...
> ploomber install
...
> ploomber nb --inject
```

jinniw43805:
## Describe your changes

This PR is to fix Fix #1048. 
The user was not able to execute `nb --inject` after initializing the empty package project with `scaffold --package`

### Problem Root Cause

https://github.com/ploomber/ploomber/blob/master/src/ploomber/cli/nb.py#L84

The function was trying to access the class `PythonCallable` with `_path` attribute. However, the `_path` member only defined in `NotebookSource`

### Fix

By setting the _path attribute in `PythonCallable` class according to the primitive type

### Test Case

Add some test cases:
- [x] PythonCallableSource with callable type in `tests/tasks/test_tasks.py`
- [x] PythonCallableSource with string path type in `tests/sources/test_sources.py`

### Description

Discussion Items (may convert to action items later):
- [ ] The definition of the path of the callable primitive type should be the file invokes the PythonCallable? e.g. Calling the `PythonCallable` API in the `test_tasks.py`, the path name should also be `test_tasks.py`?
- [ ] Should we consider making `_path` in PythonCallableSource a public member, since it was accessed & used in application level
- [x] Think about side-effects - test if [_get_default_task_to_inject_by_template_name](https://github.com/ploomber/ploomber/blob/master/src/ploomber/cli/nb.py#L67) can be executed successfully
- [x] Investigate the codebase, what scenarios are for `PythonCallableSource` and  `NotebookSource`
- [x] Need to check the codebase if Source classes are not having `_path` member
- [ ] How about stdin mode in Python CLI?

Todo Items:
- [x] Identify the problem 
- [x] Reproduce the issue in the unit testing level
- [x] Cleanup the code

[Closes #1048
](https://github.com/ploomber/ploomber/issues/1048)

### DIFF
--- a/tests/tasks/test_tasks.py
+++ b/tests/tasks/test_tasks.py
@@ -169,6 +169,7 @@ def test_python_callable_with_file():
     assert str(t.product) == 'file.txt'
     assert str(t.source) == ('def touch(product):\n    '
                              'Path(str(product)).touch()\n')
+    assert str(Path(t.source._path).name) == "test_tasks.py"
 
 
 def test_postgresscript_with_relation(pg_client_and_schema):
--- a/src/ploomber/sources/pythoncallablesource.py
+++ b/src/ploomber/sources/pythoncallablesource.py
@@ -128,7 +128,7 @@ class PythonCallableSource(Source):
         self._hot_reload = hot_reload
         self._needs_product = needs_product
         if isinstance(primitive, str):
-            # If the primitive is the string path, the path will be the file invokes the loc, will be lazily set when _loc is loaded 
+            # If the primitive is the string path, the path will be the file invokes the loc, will be lazily set when _loc is loaded
             self._path = None
         elif callable(primitive):
             self._path = self._callable_loader.get_path()
--- a/src/ploomber/sources/pythoncallablesource.py
+++ b/src/ploomber/sources/pythoncallablesource.py
@@ -82,6 +82,7 @@ class CallableLoader:
 
     def get_path(self):
         return getfile(self.load())
+
     @property
     def from_dotted_path(self):
         return self._from_dotted_path
@@ -131,6 +132,7 @@ class PythonCallableSource(Source):
             self._path = None
         elif callable(primitive):
             self._path = self._callable_loader.get_path()
+
     @property
     def hot_reload(self):
         return self._hot_reload
--- a/src/ploomber/sources/pythoncallablesource.py
+++ b/src/ploomber/sources/pythoncallablesource.py
@@ -128,7 +128,8 @@ class PythonCallableSource(Source):
         self._hot_reload = hot_reload
         self._needs_product = needs_product
         if isinstance(primitive, str):
-            # If the primitive is the string path, the path will be the file invokes the loc, will be lazily set when _loc is loaded
+            # If the primitive is the string path
+            # will be lazily set when _loc is loaded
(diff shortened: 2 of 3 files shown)


tests/tasks/test_tasks.py
```diff
@@ -169,6 +169,7 @@ def test_python_callable_with_file():
     assert str(t.product) == 'file.txt'
     assert str(t.source) == ('def touch(product):\n    '
                              'Path(str(product)).touch()\n')
+    assert str(Path(t.source._path).name) == "test_tasks.py"
```
jinniw43805, line 172: This one needs to be clarified

yafimvo (approved):
@jinniw43805 good job.

Good points. At this point, it's enough for me. what do you think? @idomic

edublancas:
I'm unsure how this addresses the issue; we're missing one unit test containing a pipeline with some Python functions as tasks and some notebooks. Then, see if it injects the cell successfully into the notebooks while ignoring the functions.

idomic:
> we're missing one unit test containing a pipeline with some Python functions as tasks and some notebooks.
Shouldn't it be via scaffold?

edublancas:
> Shouldn't it be via scaffold?

No. The user had issues with scaffold not because of scaffold itself (scaffold just creates a bunch of files); but because the scaffold template contains some Python functions in the example pipeline

idomic:
@yafim let's add the missing test and check the fix works

edublancas:
also, I don't see the benefit of adding the `_path` attribute to PythonCallable. We already have `.loc` in all tasks. An easier solution would just be to check if a task is a NotebookRunner instance, and if so, inject the cell, otherwise ignoring. this is adding extra logic that has no use

edublancas:
closing this in favor of https://github.com/ploomber/ploomber/pull/1056

PR #1060, opened 2023-01-04
File client user guide

### LINKED ISSUE #701: add a user guide on File clients
We are missing a user guide explaining how to use File clients, it should explain how to use S3 and Google Cloud.

It's also important to mention how it works (it uploads artifacts to the cloud). Another important concept to explain is how to bring external datasets that live in the cloud already since File clients only upload products generated by the pipeline.

We should also link it to this example: https://github.com/ploomber/projects/tree/master/templates/google-cloud

neelasha23:
Closes #701 


idomic:
Looks better already, just make sure the CI is passing before review.

neelasha23:
> Looks better already, just make sure the CI is passing before review.

I created the branch from the latest `master` branch but I see this test failing: `tests/io_mod/test_terminalwriter.py::test_code_highlight[with markup and code_highlight]`
Is this an intermittent issue?

@idomic @edublancas

doc/user-guide/file_clients.rst
```diff
@@ -0,0 +1,87 @@
...
+
+    This is a guide on ``file clients``. For API docs
+    see :ref:`api-python-client`.
+
+File clients are used for uploading File products to the cloud. Currently two clients are supported for Amazon S3 and Google Cloud respectively.
+
+The upload process happens in two steps:
+
+* Given a local path, the remote path for storing the file is computed. An absolute local file path of ``/path/to/project/out/data.csv`` gets translated to ``path/to/parent/out/data.csv``. Here, ``parent`` is the parent folder in the bucket to store the files.
+* Upload the file using appropriate cloud APIs.
```
edublancas, line 14: this line is probably out of place?

doc/user-guide/file_clients.rst
```diff
@@ -0,0 +1,87 @@
...
+
+    # content continues...
+
+
+Working with external datasets
+------------------------------
+
+The file clients only upload products generated by the pipeline. If you want to work with an external dataset which is in the cloud please follow the below approach:
+
+* Sql script that runs a CREATE TABLE statement to copy a public dataset.
```
edublancas, line 72: I'm a but confused about this section. I think this is documenting what to do when the dataset we want to work with is in a different place right? But I'm not following how that connects to this list

edublancas (changes requested):
am I right to assume that these docs are mostly gathering docs that we already had in other places? (that's fine, I'm just wondering where this came from and if we should remove other references to prevent duplication)

edublancas:
>  I created the branch from the latest master branch but I see this test failing: tests/io_mod/test_terminalwriter.py::test_code_highlight[with markup and code_highlight]
Is this an intermittent issue?


yeah, I saw that in an earlier CI run. We can ignore it

doc/user-guide/file_clients.rst
neelasha23: This was mentioned in the original issue: `Another important concept to explain is how to bring external datasets that live in the cloud already since File clients only upload products generated by the pipeline.`

doc/user-guide/file_clients.rst
neelasha23: > am I right to assume that these docs are mostly gathering docs that we already had in other places? (that's fine, I'm just wondering where this came from and if we should remove other references to prevent duplication)

I picked up the client file creating and pipeline configuration bits from `soopervisor` docs and some example scripts in the projects repo. The remaining details have been added based on points listed in the [original issue](https://github.com/ploomber/ploomber/issues/701). I'm not very if the last section and notes section is present elsewhere in the docs

edublancas: sounds good

doc/user-guide/file_clients.rst
edublancas: Ok, so let's remove this three points since they're confusing and replace them with:


> The file clients only upload products generated by the pipeline. If you want to work with an external dataset, you should download such a dataset in the pipeline task that uses it as input. If you need help contact us (add link to our slack)

edublancas (changes requested):

PR #1092, opened 2023-04-02
Fixed local execution in Scriptrunner

### LINKED ISSUE #1072: ScriptRunner: Cannot import modules when other directories
I am trying to run the pipeline via the ScriptRunner and am not able to import modules from other directories. 
My current project structure looks like this:
```
└── project/
    ├── common/
    │   └── lib
    ├── file1.py --> want to import common.lib here
    ├── file2.py
    └── pipeline.yaml
```
Adding the project to the sys.path fixes this issues.

Note: This behaviour is not observed with the default NotebookRunner.

mehtamohit013:
## Describe your changes
Fixed local execution for scriptrunner
Closes #1072 

### DIFF
--- a/src/ploomber/tasks/notebook.py
+++ b/src/ploomber/tasks/notebook.py
@@ -524,8 +524,8 @@ class NotebookRunner(NotebookMixin, Task):
             Support for generating output notebooks in multiple formats, see
             example above.
 
-    `nbconvert's documentation <https://nbconvert.readthedocs.io/en/latest/config_options.html#preprocessor-options>`_ 
-    """ # noqa
+    `nbconvert's documentation <https://nbconvert.readthedocs.io/en/latest/config_options.html#preprocessor-options>`_
+    """  # noqa
 
     PRODUCT_CLASSES_ALLOWED = (File,)
 
@@ -864,7 +864,7 @@ class ScriptRunner(NotebookMixin, Task):
         params=None,
         ext_in=None,
         static_analysis="regular",
-        local_execution=False,
+        local_execution=True,
     ):
         self.ext_in = ext_in
         self.local_execution = local_execution
@@ -901,11 +901,6 @@ class ScriptRunner(NotebookMixin, Task):
         if self.static_analysis == "regular":
             self.source._check_notebook(raise_=True, check_signature=False)
 
-        if self.local_execution:
-            cwd = str(self.source.loc.parent)
-        else:
-            cwd = None
-
         fd, tmp = tempfile.mkstemp(".py")
         os.close(fd)
 
@@ -916,6 +911,17 @@ class ScriptRunner(NotebookMixin, Task):
                 if c["cell_type"] == "code"
             ]
         )
+
+        if self.local_execution:
+            cwd = str(self.source.loc.parent.resolve())
+            # Adding a code cell on top of python file to
+            # support import of local module
+            code_block = "# Adding support for local execution\n"
+            code_block = code_block + "import sys\n"
+            code_block = code_block + f'sys.path.append("{cwd}")\n\n\n'
+            code = code_block + code
+        else:
+            cwd = None
 
         tmp = Path(tmp)
         tmp.write_text(code)
@@ -936,6 +942,9 @@ class ScriptRunner(NotebookMixin, Task):
 
 
 def _run_script_in_subprocess(interpreter, path, cwd):
+
+    # For python, the cwd will not get appended to the sys path
+    # Therefore added explicit codeblock above for local execution
     res = subprocess.run([interpreter, str(path)], cwd=cwd, stderr=PIPE)
 
     if res.returncode:


mehtamohit013:
@edublancas 
ScriptRunner API has a poor [documentation](https://docs.ploomber.io/en/latest/api/_modules/tasks/ploomber.tasks.ScriptRunner.html), as function parameters such as `local_execution` are not documented.

edublancas (changes requested):
instead of adding a new cell to the notebook, it'd be better to ensure that the current working directory is in the Python path, I think you can use the `env` variable in the subprocess.run: https://docs.python.org/3/library/subprocess.html#subprocess.run

and modify the PYTHONPATH variable

mehtamohit013:
#1093

edublancas:
the CI is failing, but the errors seem unrelated to the changes, let me know if you need help

mehtamohit013:
@edublancas 
The CI is failing because of an import error due to `numpydoc` on macos with `python-3.9`. This is because of the code [here](https://github.com/ploomber/ploomber/blob/dbb25ddbcee6ebbfdc5d087c95c350419ea30e9f/src/ploomber/cli/parsers.py#L257).

I don't have a mac to reproduce the error. But I can try with GitHub actions if you want.

Btw do you know why there is try and catch block there? The import is basically failing and going into an error block, which then messes with the test case

mehtamohit013:
Found the cause of the error here: https://github.com/X-DataInitiative/tick/issues/236
They changed the import from 0.7 to 0.8

mehtamohit013:
Also, while installing dependencies, pip downloads, build and looks for multiple versions, and checks for compatibility. This can be mitigated by using `conda` or maybe `mamba`  to install rather than pip, which also makes the pipeline robust (10 min to install dependencies)
For example: [numpydoc](https://github.com/ploomber/ploomber/actions/runs/4681148498/jobs/8293364453?pr=1092#step:9:2717)

In the case of the failed test case, `numpydoc` version goes to [`0.6`](https://github.com/ploomber/ploomber/actions/runs/4681148498/jobs/8293364453?pr=1092#step:9:2802) in which the import is different from what written [here](https://github.com/ploomber/ploomber/blob/dbb25ddbcee6ebbfdc5d087c95c350419ea30e9f/src/ploomber/cli/parsers.py#L257).

Also, have a look at pip statement [above](https://github.com/ploomber/ploomber/actions/runs/4681148498/jobs/8293364453?pr=1092#step:9:2802). 

TLDR: `numpydoc` have to be `>=0.8`

edublancas:
> Btw do you know why there is try and catch block there? The import is basically failing and going into an error block, which then messes with the test case

The try catch is there because numpydoc is an optional dependency. If it isn't there, it shouldn't crash, but run some default behavior.

> Also, while installing dependencies, pip downloads, build and looks for multiple versions, and checks for compatibility. This can be mitigated by using conda or maybe mamba to install rather than pip, which also makes the pipeline robust (10 min to install dependencies)

Yeah, this is a known issue. However, we cannot rely on `conda` or `mamba` for this step since all our dependencies are declared in the `setup.py` file, which only works with `pip`. Furthermore, since most people install things with `pip` we want to ensure we're replicating that process in our CI, so we detect things when they go wrong.

> TLDR: numpydoc have to be >=0.8 based on https://github.com/X-DataInitiative/tick/issues/236

Ok, so let's try upgrading pip before installing anything. We already do this in the linux tests, so we should ensure we do it in all other workflows: https://github.com/ploomber/ploomber/blob/dbb25ddbcee6ebbfdc5d087c95c350419ea30e9f/.github/workflows/ci-unit-linux.yml#L56

if that doesn't work, we can hardcode a `pip install numpydoc --upgrade` right after we install dependencies: https://github.com/ploomber/ploomber/blob/dbb25ddbcee6ebbfdc5d087c95c350419ea30e9f/.github/workflows/ci-unit-macos.yml#L91

mehtamohit013:
With pip upgrade it works

src/ploomber/tasks/notebook.py
```diff
@@ -860,7 +892,7 @@ def __init__(
         params=None,
         ext_in=None,
         static_analysis="regular",
-        local_execution=False,
+        local_execution=True,
```
edublancas, line 895: this will break compatibility. can't we keep the old False default and enable this feature?

src/ploomber/tasks/notebook.py
```diff
@@ -913,6 +940,14 @@ def run(self):
             ]
         )
 
+        if self.local_execution:
```
edublancas, line 943: I'm a bit confused about this logic. 

seems like adding the notebook's parent should be independent from `local_execution`?

Like regardless of its value we can just add `str(self.source.loc.parent.resolve())` to the env to ensure modules are importable

edublancas (changes requested):

src/ploomber/tasks/notebook.py
mehtamohit013: Yeah sure. I just thought it is more intuitive to set it to true.

src/ploomber/tasks/notebook.py
mehtamohit013: I agree. But in the documentation of NotebookRunner, it is written:
> local_execution (bool, optional) – Change the working directory to be the parent of the notebook’s source. Defaults to False. This resembles the default behavior when running notebooks interactively via Jupyter notebook

And as `ScriptRunner` has the same documentation with the same parameters, that's why I used it. But I do agree, there is no need of `local_execution` in `ScriptRunner` and by default `env` should be used.

Do you want me to remove it?

src/ploomber/tasks/notebook.py
edublancas: let's revert to the old value to keep backward compatibility

src/ploomber/tasks/notebook.py
edublancas: yeah, let's make this fix independent of the `local_execution` parameter

edublancas (changes requested):

src/ploomber/tasks/notebook.py
```diff
@@ -860,10 +888,8 @@ def __init__(
         params=None,
         ext_in=None,
         static_analysis="regular",
-        local_execution=False,
```
edublancas, line 863: looks like the parameter was deleted? I thought we'd keep it (along with the default value)

src/ploomber/tasks/notebook.py
```diff
@@ -913,6 +934,10 @@ def run(self):
             ]
         )
 
+        cwd = str(self.source.loc.parent.resolve())
+        orig_env = os.environ.copy()
+        orig_env["PYTHONPATH"] = cwd
```
edublancas, line 939: this will overwrite any existing value, we should append the new value if PYTHONPATH exists

tests/tasks/test_notebook.py
```diff
@@ -1383,3 +1383,31 @@ def test_notebook_add_parameters_cell_if_missing(
         else jupytext.guess_format(path.read_text(), ext=path.suffix)
     )
     assert fmt == expected_fmt
+
+
+def test_local_import_ScriptRunner(tmp_directory, capsys):
```
edublancas, line 1388: let's add one more tests where PYTHONPATH is added to ensure we're not overwriting it

edublancas (changes requested):

src/ploomber/tasks/notebook.py
mehtamohit013: > yeah, let's make this fix independent of the local_execution parameter

What will the functionality of the `local_execution` parameter? Because earlier you mention, the local path will be default inserted into `PYTHONPATH`

edublancas: yeah, I agree there's some overlap. but local_execution is still different since if moves the current working directory.

I don't think many people use this option but it's better not to break the API unless we have a strong reason to.

mehtamohit013: >yeah, I agree there's some overlap. but local_execution is still different since if moves the current working directory.

I am still confused. What will happen, if I set it to `True`? What should I append to the `PYTHONPATH`? Because the script local dir is already getting appended to.

```python
orig_env["PYTHONPATH"] += ':'+"what_will_be_here"
```

mehtamohit013:
@idomic Yes, it is fixed. This PR is ready to merge

PR #1093, opened 2023-04-04
Fix tests

mehtamohit013:
Closes #x 

- [X] I have added thorough tests (when necessary).

----
:books: Documentation preview :books:: https://ploomber--1093.org.readthedocs.build/en/1093/


mehtamohit013:
@edublancas @idomic 
I have found the error and it is mainly due to recent changes to the ploomber engine [#67](https://github.com/ploomber/ploomber-engine/pull/67)

There are a few things I am unable to understand:

1. What is the role of `**globals` ,  `**kwargs` and `**upstream` in the ploomber engine shell [code](https://github.com/ploomber/ploomber/blob/2f8dd99a6c62b1032fa4927a8c75c8af6aba69bb/src/ploomber/micro/_capture.py#L164)?
2. Why there is a separate error-catching block for `"return outside function"`  [here](https://github.com/ploomber/ploomber/blob/2f8dd99a6c62b1032fa4927a8c75c8af6aba69bb/src/ploomber/micro/_capture.py#L136)
3. Why do the notebook cells keep on executing even if there is an error ([code](https://github.com/ploomber/ploomber/blob/2f8dd99a6c62b1032fa4927a8c75c8af6aba69bb/src/ploomber/micro/_capture.py#L117))?

I can temporarily fix the `_Client` class, but it is not a robust way to use ploomber-engine internal functions outside of the library, as it can break at any moment. The main culprit is my previous PR, in which I add the error cells above the failed cell, and because of this, the class is stuck in an endless `for` loop.

I have also added the commented block for a better way, but currently, ploomber engine doesn't support a way to pass parameters to the initialized shell. So, there will be no support for `**globals`, `**kwargs` and `**upstream`

edublancas:
the _micro module was a POC that didn't have success.

I think we should deprecate this module

edublancas:
I'm assuming this PR is looking to fix our CI?

the `_capture.py` [module](https://github.com/ploomber/ploomber/blob/master/src/ploomber/micro/_capture.py) was an experiment, we didn't make this public so no one should be using it. We can deprecate it.

we need to also remove this two references, there's a capture import here:
https://github.com/ploomber/projects/blob/master/micro-pipelines/2-guide.ipynb
and we need to also delete this file:
https://github.com/ploomber/projects/blob/master/micro-pipelines/capture.py

mehtamohit013:
> I'm assuming this PR is looking to fix our CI?

Yes

> the `_capture.py` [module](https://github.com/ploomber/ploomber/blob/master/src/ploomber/micro/_capture.py) was an experiment, we didn't make this public so no one should be using it. We can deprecate it.

> we need to also remove this two references, there's a capture import here: https://github.com/ploomber/projects/blob/master/micro-pipelines/2-guide.ipynb and we need to also delete this file: https://github.com/ploomber/projects/blob/master/micro-pipelines/capture.py

Ok, I will remove the module and any reference of it in ploomber

mehtamohit013:
I have currently put `pytest.xfail` decorator around the test that uses the `_capture` module.
Let me know if any test is essential and needs to be modified.

src/ploomber/micro/_capture.py
```diff
@@ -21,8 +21,7 @@
```
edublancas, line 21: since this module is deprecated, we can delete the whole file

tests/test_micro.py
```diff
@@ -187,6 +187,7 @@ def second(first):
     dag.build()
 
 
+@pytest.mark.xfail(reason="_capture module has been deprecated")
```
edublancas, line 190: once we delete the _micro file, we'll see some tests failing, we should delete them instead of marking them with xfail

edublancas (changes requested):

mehtamohit013:
[tests/sources/test_python_interact.py](https://github.com/ploomber/ploomber/pull/1093/files#diff-73b3aa3542e7f300c1ec2c1cea4d497d52ee10452a6dc252c838e7f47b4beec6) has only different formatting.

edublancas (approved):

PR #1101, opened 2023-04-26
Add support for ploomber engine

mehtamohit013:
## Issue number

Closes #1082 #1091 

## Checklist before requesting a review

- [X] Performed a self-review of my code
- [X] Formatted my code with [`pkgmt format`](https://ploomber-contributing.readthedocs.io/en/latest/contributing/pr.html#linting-formatting)
- [X] Added [tests](https://ploomber-contributing.readthedocs.io/en/latest/contributing/pr.html#testing) (when necessary).
- [X] Added [docstring](https://ploomber-contributing.readthedocs.io/en/latest/contributing/pr.html#documenting-changes-and-new-features) documentation and update the [changelog](https://ploomber-contributing.readthedocs.io/en/latest/contributing/pr.html#changelog) (when needed)

----
:books: Documentation preview :books:: https://ploomber--1101.org.readthedocs.build/en/1101/

PR #1123, opened 2023-07-07
Cleanup tests

### LINKED ISSUE #1109: clean up tests and better error message with R notebooks
we need to address a few things from this PR: https://github.com/ploomber/ploomber/pull/1101

1. refactor unit tests: https://github.com/ploomber/ploomber/pull/1101#discussion_r1212466632
2. better error message when trying to run an R notebook with ploomber-engine https://github.com/ploomber/ploomber/pull/1101#discussion_r1211999094

mehtamohit013:
## Issue number

Closes #1109 

- [X] Performed a self-review of my code
- [X] Formatted my code with [`pkgmt format`](https://ploomber-contributing.readthedocs.io/en/latest/contributing/pr.html#linting-formatting)
- [X] Added [tests](https://ploomber-contributing.readthedocs.io/en/latest/contributing/pr.html#testing) (when necessary).
- [X] Added [docstring](https://ploomber-contributing.readthedocs.io/en/latest/contributing/pr.html#documenting-changes-and-new-features) documentation and update the [changelog](https://ploomber-contributing.readthedocs.io/en/latest/contributing/pr.html#changelog) (when needed)

----
:books: Documentation preview :books:: https://ploomber--1123.org.readthedocs.build/en/1123/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -104,6 +104,7 @@ TESTING = [
     "dill==0.3.5.1",
     # pandas not yet compatible with sqlalchemy 2
     "sqlalchemy<2",
+    "ipython_genutils",
 ]
 
 # packages needed for development
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ def read(name):
 # https://github.com/nteract/papermill/issues/239
 NB = [
     "papermill",
-    "notebook<=6.5.4",
+    "notebook<7",
     "jupytext",
     "ipykernel>=1.5.2",
     "jupyter_client>=5.3.1",
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,7 @@ def read(name):
 # https://github.com/nteract/papermill/issues/239
 NB = [
     "papermill",
+    "notebook<=6.5.4",
     "jupytext",
     "ipykernel>=1.5.2",
     "jupyter_client>=5.3.1",
@@ -104,7 +105,6 @@ TESTING = [
     "dill==0.3.5.1",
     # pandas not yet compatible with sqlalchemy 2
     "sqlalchemy<2",
-    "ipython_genutils",
 ]
 
 # packages needed for development
--- a/src/ploomber/util/dotted_path.py
+++ b/src/ploomber/util/dotted_path.py
@@ -476,7 +476,10 @@ class DottedPathSpecModel(BaseModel):
             )
 
     def get_kwargs(self):
-        return self.dict(exclude={"dotted_path"})
+        if int(pydantic.__version__.split(".")[0]) >= 2:
+            return self.model_dump(exclude={"dotted_path"})
+        else:
+            return self.dict(exclude={"dotted_path"})
 
 
 def create_intermediate_modules(module_parts):
--- a/tests/tasks/test_notebook.py
+++ b/tests/tasks/test_notebook.py
@@ -194,8 +194,8 @@ does_not_raise = MyNullContext()
             "sample.R",
             ".",
             "ploomber-engine",
-            pytest.raises(DAGBuildError),
-            "NameError: name 'c' is not defined",
+            pytest.raises(ValueError),
+            "Ploomber Engine currently does not support R notebooks",
         ],
     ],
 )
@@ -204,14 +204,15 @@ def test_execute_sample_nb_R(
 ):
     dag = DAG()
 
-    NotebookRunner(
-        Path(name),
-        product=File(Path(out_dir, name + ".out.ipynb")),
-        dag=dag,
-        executor=executor,
-    )
     with expectation as excinfo:
+        NotebookRunner(
+            Path(name),
+            product=File(Path(out_dir, name + ".out.ipynb")),
+            dag=dag,
+            executor=executor,
+        )
         dag.build()
+
         if excinfo:
             assert error_msg in str(excinfo.value)
 
--- a/tests/tasks/test_notebook.py
+++ b/tests/tasks/test_notebook.py
@@ -20,7 +20,6 @@ from ploomber.exceptions import (
 from ploomber.executors import Serial
 from ploomber.products import File
 from ploomber.sources.nb_utils import find_cell_with_tag
-from ploomber.sources import NotebookSource
 from ploomber.tasks import NotebookRunner, notebook, ScriptRunner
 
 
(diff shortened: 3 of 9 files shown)


mehtamohit013:
> refactor unit tests: https://github.com/ploomber/ploomber/pull/1101#discussion_r1212466632

Done

>better error message when trying to run an R notebook with ploomber-engine Add support for ploomber engine #1101 

So, what is happening here, is there is no filter in `ploomber-engine` to detect the file extension type, so it always assume it is `python` and pass the code to `InteractiveShell` of `Ipython`. In this case the code passed is ` c(1, 2, 3)`, which results in `NameError: name 'c' is not defined`. The ideal fix, I think would be to add file language detection to `ploomber-engine`

mehtamohit013:
@edublancas 
Test Failing:
I believe something changed in `black` in `pkgmt`. I am not quite sure, how to resolve it. Any ideas?

Error:
```
error: cannot format /home/runner/work/ploomber/ploomber/src/ploomber/resources/ploomber_add/function.py: Cannot parse: 1:4: def [[function_name]](product):
error: cannot format /home/runner/work/ploomber/ploomber/src/ploomber/resources/ploomber_add/task.py: Cannot parse: 11:1: [% if not is_ipynb -%]
```

edublancas:
the files in `ploomber_add` are jinja templates (hence, they have invalid python code). the problem is that flake8 is trying to parse them.

flake8 is configured to ignore those files: https://github.com/ploomber/ploomber/blob/7c11ac7197de89add9909e60eb2069632075ac8c/setup.cfg#L8 so I'm unsure why it's taking them into account. I looked at pypi.org and there hasn't been any recent flake8 releases so more investigation is needed

mehtamohit013:
> the files in `ploomber_add` are jinja templates (hence, they have invalid python code). the problem is that flake8 is trying to parse them.
> 
> flake8 is configured to ignore those files:
> 
> https://github.com/ploomber/ploomber/blob/7c11ac7197de89add9909e60eb2069632075ac8c/setup.cfg#L8
> so I'm unsure why it's taking them into account. I looked at pypi.org and there hasn't been any recent flake8 releases so more investigation is needed

I believe it is `black` which is failing . 

Weird behaviour:
So, `black .` is passing, but `black . --extend-exclude ''` is failing, `black . --exclude ''` is passing.
Relevant PR: https://github.com/ploomber/pkgmt/pull/62

@edublancas

edublancas:
looks like something broke in the latest pkgmt version. then let's fix it and then continue here

mehtamohit013:
@edublancas 
```
========================== short test summary info ============================
FAILED tests/spec/test_taskspec.py::test_process_dotted_paths[valid-dotted-path-str]
FAILED tests/spec/test_taskspec.py::test_process_dotted_paths[valid-dotted-path-dict]
FAILED tests/util/test_dotted_path.py::test_call_spec_without_dotted_path_key
FAILED tests/util/test_dotted_path.py::test_dotted_path_if_overriding_args - ...
= 4 failed, 2781 passed, 11 skipped, 6 xfailed, 1 xpassed, 561 warnings in 487.91s (0:08:07) =
```

@edublancas 
Some tests are failing which are not related to this PR and have not been modified, is there any open issue or PR related to these?

edublancas:
there's no open issue. did you checked the error messages? usually these things are due to breaking changes in the dependencies. so we might have to either support a new version or pin it. but we'd have to figure out what's the root cause first

mehtamohit013:
@edublancas 
It seems like `taskspec` uses `pydantic` internally, and `pydantic` `V2` introduced some API breaking changes. So do you want me to pin `V1` or migrate to `V2`?

[Migration Guide](https://docs.pydantic.dev/dev-v2/migration/)

edublancas:
is the only problem calling `dict`? if so, then we can add a conditional that calls dict if the user has pydantic 1 installed or model_dump otherwise. If the migration is more difficult then I think it's fine to migrate

mehtamohit013:
Btw, the testing contains some tests, which fail sometime or pass time due to unknown errors:

Encountered:
linux python-3.8: `tests/products/test_file.py::test_task_with_client_and_metaproduct_with_some_missing_products`
windows python-3.9: `tests/sources/test_python_interact.py::test_signature_line_break`

mehtamohit013:
@edublancas 
Ready for review and merge

src/ploomber/util/dotted_path.py
```diff
@@ -476,7 +476,10 @@ def from_spec(cls, dotted_path_spec, strict=False):
             )
 
     def get_kwargs(self):
-        return self.dict(exclude={"dotted_path"})
+        if int(pydantic.__version__.split(".")[0]) >= 2:
```
edublancas, line 479: add an entry to the changelog to specify that we now support pydantic 2.0

tests/tasks/test_notebook.py
```diff
@@ -167,124 +166,79 @@ def test_notebook_conversion_stores_as_unicode(tmp_directory, monkeypatch):
...
-        ["sample.R", ".", "papermill"],
-        ["sample.R", ".", "ploomber-engine"],
-        ["sample.R", ".", "some-other-executor"],
+        ["sample.R", ".", "papermill", does_not_raise, None],
+        [
+            "sample.R",
+            ".",
+            "ploomber-engine",
+            pytest.raises(DAGBuildError),
+            "NameError: name 'c' is not defined",
```
edublancas, line 198: part of the AC was to improve the error message when trying to run an R notebook with ploomber-engine, has this been implemented? 

https://github.com/ploomber/ploomber/issues/1109#issuecomment-1600642210

edublancas (changes requested):

tests/tasks/test_notebook.py
mehtamohit013: [Earlier Comment](https://github.com/ploomber/ploomber/pull/1123#issuecomment-1625590595)
> So, what is happening here, is there is no filter in ploomber-engine to detect the file extension type, so it always assume it is python and pass the code to InteractiveShell of Ipython. In this case the code passed is  c(1, 2, 3), which results in NameError: name 'c' is not defined. The ideal fix, I think would be to add file language detection to ploomber-engine

src/ploomber/util/dotted_path.py
mehtamohit013: Added Changelog

tests/tasks/test_notebook.py
edublancas: yes, so part of the AC for this issue is precisely to improve the error message. Inside NotebookRunner, you can access the source property, which will return a NotebookSource (unsure if this is the exact name), I think there is a property there that return's the notebook's language. then, if you find it's an R notebook, you can raise an error if the user tries to use ploomber-engine

edublancas (changes requested):
the only missing part is improving the error message when trying to run an R notebook

mehtamohit013:
Currently Failing: 

```python
ImportError while importing test module '/home/runner/work/ploomber/ploomber/tests/test_jupyter.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/share/miniconda/envs/conda-env/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_jupyter.py:11: in <module>
    from ipython_genutils.tempdir import TemporaryDirectory
E   ModuleNotFoundError: No module named 'ipython_genutils'
```

Working on it..

edublancas:
@mehtamohit013 yes, makes sense. let's pin it. once we're done, we can check support for notebook 7

setup.py
```diff
@@ -31,6 +31,7 @@ def read(name):
 # https://github.com/nteract/papermill/issues/239
 NB = [
     "papermill",
+    "notebook<=6.5.4",
```
edublancas, line 34:
```suggestion
    "notebook<7",
```

better to pin this way, because new notebook 6.x versions will come out

edublancas (changes requested):
just a minor change

edublancas (approved):

PR #1176, opened 2025-06-04
allow disabling of static_analysis

### LINKED ISSUE #1175: Ploomber expects all parameters in config file to be used in notebooks run with NotebookRunner
If global parameters are not used, ploomber prints off numerous warnings that seem concerning.

We are using a configuration file to run numerous notebooks, and not every notebook uses every global parameter in the parameter-tagged cell. 

These are some of the resulting warnings:

Image

We'd like to have an option to disable these warnings, eg during the instantiation of the NotebookRunner.

TeaganKing:
## Describe your changes
This change would allow users to specify that they would like to disable unnecessary warnings related to the global parameters not being used in every notebook run with NotebookRunner. Note that `disable` is already a [valid option for `static_analysis`](https://github.com/TeaganKing/ploomber/blob/536c8265b024a0b152745d268ec7b540e9f11ed9/src/ploomber/sources/notebooksource.py#L251).

Closes #1175

----
📚 Documentation preview 📚: https://ploomber--1176.org.readthedocs.build/en/1176/

### DIFF
--- a/.github/workflows/ci-unit-linux.yml
+++ b/.github/workflows/ci-unit-linux.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10']
+        python-version: ['3.11', '3.12']
     
     services:
       postgres:
--- a/.github/workflows/ci-unit-macos.yml
+++ b/.github/workflows/ci-unit-macos.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10']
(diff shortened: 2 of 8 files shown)


TeaganKing:
The unit test that failed has a `black` error in  `/Users/runner/work/ploomber/ploomber/tests/io_mod/test_wcwidth.py` was a change that seemed irrelevant to this PR and also less than ideal to make. Thus, I also added a `# fmt` change on those two lines in `test_wcwidth.py`.

TeaganKing:
Hi @edublancas ,  this should be ready for review!

edublancas:
Hi @TeaganKing: I tried fixing the tests but encountered many issues. most likely because of outdated dependencies. Unfortunately, I don't have the bandwidth to work on this right now. I recommend forking the project and applying your changes there.

